PR #3, opened 2017-03-14
DBInterface supports redis connection options

qiluo-msft:
Related to https://github.com/Azure/SONiC/issues/51
Trying to solve in below steps:
- sonic-py-swsssdk
- sonic-snmpagent

### DIFF
--- a/src/sswsdk/dbconnector.py
+++ b/src/sswsdk/dbconnector.py
@@ -9,6 +9,9 @@ from .interface import DBInterface
 
 
 class SonicV1Connector(DBInterface):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
     pass
 
 
@@ -19,6 +22,9 @@ if len(SonicV1Connector.db_map) != len({v['db'] for k, v in SonicV1Connector.db_
 
 
 class SonicV2Connector(DBInterface):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
     pass
 
 
--- a/src/sswsdk/dbconnector.py
+++ b/src/sswsdk/dbconnector.py
@@ -10,7 +10,7 @@ from .interface import DBInterface
 
 class SonicV1Connector(DBInterface):
     def __init__(self, **kwargs):
-        super(SonicV2Connector, self).__init__(**kwargs)
+        super(SonicV1Connector, self).__init__(**kwargs)
 
     pass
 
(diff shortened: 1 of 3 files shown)


src/sswsdk/dbconnector.py
```diff
@@ -9,6 +9,9 @@
 
 
 class SonicV1Connector(DBInterface):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
```
pavel-shirshov: https://docs.python.org/2/library/functions.html#super
class C(B):
    def method(self, arg):
        super(C, self).method(arg) # I used to such call of the parent constructor. I'm not sure will super() work for python 2.

src/sswsdk/interface.py
```diff
@@ -134,10 +134,15 @@ class DBInterface(object):
...
-    def __init__(self):
+    def __init__(self, **kwargs):
 
         super(DBInterface, self).__init__()
 
+        # Store the arguments for redis client
+        self.redis_kwargs = kwargs
+        if len(self.redis_kwargs) == 0:
+            self.redis_kwargs['unix_socket_path'] = self.REDIS_UNIX_SOCKET_PATH
+
```
pavel-shirshov: Are you not going to do anything if the length not equal to 0?
Are you sure that 'unix_socket_path' exists in the parameter?

src/sswsdk/util.py
```diff
@@ -12,20 +12,26 @@
...
+    options, remainders = getopt(sys.argv[1:], "t:p:s:d:f:h", ["host=", "port=", "unix_socket_path=", "debug=", "frequency=", "help"])
 
     args = {}
     for (opt, arg) in options:
         if opt in ('-d', '--debug'):
             args['log_level'] = int(arg)
+        elif opt in ('-t', '--host'):
+            args['host'] = arg
+        elif opt in ('-p', '--port'):
+            args['port'] = int(arg)
```
pavel-shirshov: What if arg is not a number?
Probably we need to catch exceptions here?

src/sswsdk/dbconnector.py
jleveque: Pavel is correct. `super(C, self)` is compatible with both Python 2 & 3. `super()` is available in Python 3 only.

src/sswsdk/interface.py
qiluo-msft: I will not change it if the kwargs is non empty.
The [redis client](https://redis-py.readthedocs.io/en/latest/#redis.StrictRedis) will smartly handle all the kwargs. If no unix_socket_path provided, it will open a network socket, either according to kwargs or default values

PR #6, opened 2017-04-28
Fix memory leak. Subscribe on key changes only when we need it

pavel-shirshov:
Previously database connector subscribed on key notifications, but didn't read messages from publisher until there was no any issue with reading data values. So output queues for clients which use this library was getting bigger and bigger until redis closed data connection.
With this patch I subscribe on key changes only when we waiting for data.
It works in following way:
1. Getting no data exception.
2. Subscribe to key updates
3. Retry request
4. if getting the error again, blocking and waiting for updates.
5. if there're updates, retry request
6. if there're no updates reraise exception
7. after this unsubscribe from key updates
8. and return the data to requestor.

Previously there was error, if there no data, forever, the request would have been performed for ever.

### DIFF
--- a/src/swsssdk/interface.py
+++ b/src/swsssdk/interface.py
@@ -234,7 +234,7 @@ class DBInterface(object):
         logger.debug("Subscribe to keyspace notification")
         client = self.redis_clients[db_name]
         pubsub = client.pubsub()
-        pubsub.psubscribe(self.KEYSPACE_PATTERN) # FIXME: subscribe to specific pattern
+        pubsub.psubscribe(self.KEYSPACE_PATTERN)
         self.keyspace_notification_channels[db_name] = pubsub
 
     def _unsubscribe_keyspace_notification(self, db_name):
--- a/src/swsssdk/interface.py
+++ b/src/swsssdk/interface.py
@@ -280,7 +280,7 @@ class DBInterface(object):
         client = self.redis_clients[db_name]
         val = client.hget(_hash, key)
         if not val:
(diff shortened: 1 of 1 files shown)


lguohan:
>             raise UnavailableDataError(message, key)

I think this should be _hash, not key, because in the keyspace notification message, the msg.get('data') should be the redis _hash.

---
Refers to: src/swsssdk/interface.py:285 in 440d988. [](commit_id = 440d988ce01301855a8f632681662cd631766999, deletion_comment = False)

lguohan:
>             message = "Key '{}' unavailable in database '{}' - table '{}'".format(key, _hash, db_name)

the message should be 

"Key '{}' field unavailable in database '{}'".format(_hash, key, database)

---
Refers to: src/swsssdk/interface.py:283 in 440d988. [](commit_id = 440d988ce01301855a8f632681662cd631766999, deletion_comment = False)

lguohan:
>             message = "Table '{}' does not exist in database '{}'".format(_hash, db_name)

"Key '{}' unavailable in database '{}'".format(_hash, database)

---
Refers to: src/swsssdk/interface.py:301 in 440d988. [](commit_id = 440d988ce01301855a8f632681662cd631766999, deletion_comment = False)

qiluo-msft:
>         """

Add comment for return value.

---
Refers to: src/swsssdk/interface.py:323 in 38ff1e2. [](commit_id = 38ff1e2c500338854cac66247eceabdde46b2594, deletion_comment = False)

PR #8, opened 2017-07-18
Add support to subscribe to config DB change

taoyl-ms:


src/swsssdk/configdb.py
```diff
@@ -2,29 +2,83 @@
...
-        """Connect to Redis through TCP, which does not requires root.
-        """
+        # Connect to Redis through TCP, which does not requires root.
         super(ConfigDBConnector, self).__init__(host='127.0.0.1')
+        self.handlers = {}
 
     def connect(self):
         SonicV2Connector.connect(self, self.CONFIG_DB, False)
 
+    def add_handler(self, table, handler):
```
lguohan: maybe rename to subscribe(self, table, handler) is more common.

I am not sure if we need to add multiple handers to a same table. Do we have such use caes? if we do not need to, then the api signature becomes simpler

subscribe(table, handler)
unsubscribe(table)

src/swsssdk/configdb.py
```diff
@@ -99,4 +153,6 @@ def get_config(self):
                 data[table_name] = {}
             data[table_name][key] = client.hgetall(_hash)
         return data
+        self.pubsub = self.redis_clients[self.CONFIG_DB].pubsub()
+        self.pubsub.psubscribe("__keyspace@{}__:*".format(self.db_map[self.CONFIG_DB]['db']))
```
lguohan: how can these code be reached?

src/swsssdk/configdb.py
```diff
@@ -2,29 +2,83 @@
...
+    def listen(self):
+        """Start listen Redis keyspace events and will trigger corresponding handlers when content of a table changes.
+        """
+        self.pubsub = self.redis_clients[self.CONFIG_DB].pubsub()
+        self.pubsub.psubscribe("__keyspace@{}__:*".format(self.db_map[self.CONFIG_DB]['db']))
+        for item in self.pubsub.listen():
+            if item['type'] == 'pmessage':
+                _hash = item['channel'].split(':', 1)[1]
+                table = _hash.split(':', 1)[0]
+                key = _hash.split(':', 1)[1]
```
lguohan: in case someone just set a key in the config_db, here you won't get the right message, your listener will crash. It is better to ignore those messages and not to crash.

src/swsssdk/configdb.py
```diff
@@ -2,29 +2,83 @@
...
+                if self.handlers.has_key(table):
+                    client = self.redis_clients[self.CONFIG_DB]
+                    data = client.hgetall(_hash)
+                    self.__fire(table, key, data)
```
lguohan: here you are sending all hash data to the callback. I wonder if we should only send the changes to the user. All differ the set event versus del event.

lguohan:
:clock1:

src/swsssdk/configdb.py
taoyl-ms: Deleted.

---
In reply to: [128202084](https://github.com/Azure/sonic-py-swsssdk/pull/8#discussion_r128202084) [](ancestors = 128202084)

src/swsssdk/configdb.py
taoyl-ms: I think it is better to send all data to user and let user do the compare. Or else, in case that user needs other data besides changed data to do handling, both user and sdk will need to maintain a cache of table, which is a waste of effort.

---
In reply to: [128207346](https://github.com/Azure/sonic-py-swsssdk/pull/8#discussion_r128207346) [](ancestors = 128207346)

src/swsssdk/configdb.py
```diff
@@ -2,28 +2,83 @@
...
+        """Set a handler to handle config change in certain table.
+        Note that a single handler can be registered to different tables by 
+        calling this fuction multiple times.
+        Args:
+            table: Table name.
+            handler: a handler function that has signature of handler(table_name, key, data)
+        """
+        self.handlers[table] = handler
+
+    def remove_handler(self, table):
```
lguohan: -> unsubscribe?

src/swsssdk/configdb.py
```diff
@@ -93,10 +150,12 @@ def get_config(self):
         hashes = client.keys('*')
         data = {}
         for _hash in hashes:
-            table_name = _hash.split(':', 1)[0]
-            key = _hash.split(':', 1)[1]
-            if not data.has_key(table_name):
-                data[table_name] = {}
-            data[table_name][key] = client.hgetall(_hash)
+            tokens = _hash.split(':', 1)
+            if len(tokens) == 2:
```
lguohan: if number of tokens is not equal to 2, should we print out warning message? as we are not suppose to get such data from configdb?

taoyl-ms: We do have such data in configdb. One current example is the initialization state of configdb. We might later add similar configdb-internal data that we don't want to be directly readable by user.

taoyl-ms: We do have such data in configdb. One current example is the initialization state of configdb. We might later add similar configdb-internal data that we don't want to be directly readable by user.

---
In reply to: [129188316](https://github.com/Azure/sonic-py-swsssdk/pull/8#discussion_r129188316) [](ancestors = 129188316)

src/swsssdk/configdb.py
```diff
@@ -2,28 +2,106 @@
...
         'admin_status': state
         })
 
-TODO:
-    Will add API on keyspace subscription for incremental configuration support
-    in the incoming versions.
+    # Daemon to watch config change in certain table:
+    config_db = ConfigDBConnector()
+    handler = lambda table, key, data: print (key, data)
+    config_db.add_handler('BGP_NEIGHBOR', handler)
```
lguohan: subscribe?

src/swsssdk/configdb.py
```diff
@@ -2,28 +2,106 @@
...
+                        client = self.redis_clients[self.CONFIG_DB]
+                        data = self.__raw_to_typed(client.hgetall(_hash))
+                        self.__fire(table, key, data)
+
+    def __raw_to_typed(self, raw_data):
+        if raw_data == None:
+            return None
+        typed_data = {}
+        for key in raw_data:
+            if key.endswith("@"):
```
lguohan: can you add a comment and todo here?

src/swsssdk/configdb.py
```diff
@@ -93,10 +173,12 @@ def get_config(self):
...
-            if not data.has_key(table_name):
-                data[table_name] = {}
-            data[table_name][key] = client.hgetall(_hash)
+            tokens = _hash.split(':', 1)
+            if len(tokens) == 2:
+                table_name = _hash.split(':', 1)[0]
+                key = _hash.split(':', 1)[1]
+                if not data.has_key(table_name):
+                    data[table_name] = {}
+                data[table_name][key] = self.__raw_to_typed(client.hgetall(_hash))
```
lguohan: what if _hash is deleted before you call hgetall? can the code behave?

src/swsssdk/configdb.py
```diff
@@ -93,10 +173,12 @@ def get_config(self):
...
         for _hash in hashes:
-            table_name = _hash.split(':', 1)[0]
-            key = _hash.split(':', 1)[1]
-            if not data.has_key(table_name):
-                data[table_name] = {}
-            data[table_name][key] = client.hgetall(_hash)
+            tokens = _hash.split(':', 1)
+            if len(tokens) == 2:
+                table_name = _hash.split(':', 1)[0]
+                key = _hash.split(':', 1)[1]
```
lguohan: suggest to use try except block here

```
entry = self.__raw_to_typed(client.hgetall(_hash))
try:
   (table_name, key) = _hash.split(':', 1)
   if not data.has_key(table_name)
       data[table_name] = {}
   data[table_name][key] = entry
except:
    pass
```

src/swsssdk/configdb.py
```diff
@@ -63,7 +141,9 @@ def get_table(self, table):
         keys = client.keys(pattern)
         data = {}
         for key in keys:
-            data[key.split(':')[1]] = client.hgetall(key)
+            tokens = key.split(':', 1)
```
lguohan: use try except block make the code more readable.

src/swsssdk/configdb.py
```diff
@@ -2,28 +2,106 @@
...
-        """
+        # Connect to Redis through TCP, which does not requires root.
         super(ConfigDBConnector, self).__init__(host='127.0.0.1')
+        self.handlers = {}
+
+    def __wait_for_db_init(self):
+        client = self.redis_clients[self.CONFIG_DB]
+        initialized = client.get('CONFIG_DB_INITIALIZED')
+        while not initialized:
+            time.sleep(.1)
```
lguohan: why not use subscribe mechanism here instead of polling here?

src/swsssdk/configdb.py
taoyl-ms: The entry of deleted key will be returned as None. Do you have any concern regarding that?

---
In reply to: [129326056](https://github.com/Azure/sonic-py-swsssdk/pull/8#discussion_r129326056) [](ancestors = 129326056)

lguohan: i think it is better to be {} for the application code to process.

---
In reply to: [129395275](https://github.com/Azure/sonic-py-swsssdk/pull/8#discussion_r129395275) [](ancestors = 129395275,129326056)

PR #12, opened 2017-08-29
Support hierarchical keys

taoyl-ms:
1. Use `|` instead of `:` for separator in redis key, and support hierarchical multiple keys.
2. Support object with no attribute (`data == {}`) with a `"NULL": "NULL"` placeholder. 
3. `set_entry` now support delete entry with `data == None`.

### DIFF
--- a/src/swsssdk/configdb.py
+++ b/src/swsssdk/configdb.py
@@ -25,6 +25,8 @@ from .dbconnector import SonicV2Connector
 class ConfigDBConnector(SonicV2Connector):
 
     INIT_INDICATOR = 'CONFIG_DB_INITIALIZED'
+    TABLE_NAME_SEPARATOR = '|'
+    KEY_SEPARATOR = '|'
 
     def __init__(self):
         # Connect to Redis through TCP, which does not requires root.
@@ -85,7 +87,7 @@ class ConfigDBConnector(SonicV2Connector):
             if item['type'] == 'pmessage':
                 key = item['channel'].split(':', 1)[1]
                 try:
-                    (table, row) = key.split('|', 1)
+                    (table, row) = key.split(self.TABLE_NAME_SEPARATOR, 1)
                     if self.handlers.has_key(table):
                         client = self.redis_clients[self.CONFIG_DB]
                         data = self.__raw_to_typed(client.hgetall(key))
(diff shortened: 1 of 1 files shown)


src/swsssdk/configdb.py
```diff
@@ -84,7 +85,7 @@ def listen(self):
             if item['type'] == 'pmessage':
                 key = item['channel'].split(':', 1)[1]
                 try:
-                    (table, row) = key.split(':', 1)
+                    (table, row) = key.split('|', 1)
```
lguohan: >'|' [](start = 45, length = 3)

can you use a variable?

lguohan:
as a general comment, we need unit test for this configdb class.

src/swsssdk/configdb.py
```diff
@@ -117,16 +123,30 @@ def __typed_to_raw(self, typed_data):
                 raw_data[key] = value
         return raw_data
 
+    @staticmethod
+    def serialize_key(key):
+        string_types = str if sys.version_info[0] == 3 else basestring
```
lguohan: >string_types = str if sys.version_info[0] == 3 else basestring [](start = 8, length = 62)

in general, I found such shortcut conditional not quite readable, I prefer traditional if else condition.

src/swsssdk/configdb.py
```diff
@@ -117,16 +123,30 @@ def __typed_to_raw(self, typed_data):
...
+        return key if isinstance(key, string_types) else '|'.join(key)
+
+    @staticmethod
+    def deserialize_key(key):
+        tokens = key.split('|')
+        return tuple(tokens) if len(tokens) > 1 else tokens[0]
```
lguohan: >eturn tuple(tokens) if len(tokens) > 1 else tokens[0] [](start = 9, length = 53)

not readable, I think we should try to avoid to keep the code more readable.

lguohan:
>             { 'row_key': {'column_key': 'value', ...}, ...}

can you add hierachical key examples? this will be one of most heavily used library, should be heavily documented.

---
Refers to: src/swsssdk/configdb.py:171 in f975305. [](commit_id = f9753052f3901d68a567309febe1f4c8f5c16520, deletion_comment = False)

taoyl-ms:
Agree. Will add in later PRs.

---
In reply to: [326614104](https://github.com/Azure/sonic-py-swsssdk/pull/12#issuecomment-326614104) [](ancestors = 326614104)

PR #14, opened 2017-09-26
Common utility functions for bridge/port mapping

prsunny:
The port-mapping functionality is common for SNMP and FDB show CLI implementation. Moving this functionality currently in SNMP module to common library so that this can be reused by both the components.


prsunny:
Once this is merged, I shall update the SNMP module and FDB show module.

PR #15, opened 2017-09-30
Add support to remove the key of table entry

liuqu:
The class ConfigDBConnector has no method to remove the column_key of table entry.  To support it, the method "set_entry" is modified to delete the column_key which value is None.


lguohan:
merged #23, we do not explicit del the unwanted fields with the set_entry.

PR #17, opened 2017-10-17
[port_util]: Return empty dictionary instead of a NoneType in get_bridge_port_map

stcheng:
This will prevent snmp subagent from crashing when iterating an empty bridge port map

Signed-off-by: Shu0T1an ChenG <shuche@microsoft.com>


stcheng:
thanks

PR #18, opened 2017-10-25
Remove prefix "oid:0x" from sai oid

prsunny:
With a recent change in COUNTERS_DB, the SAI OIDs are prefixed with "oid:0x". However, the current scripts/utils are indexing this using key without this prefix. 

Changing to reflect the mapping as before.

### DIFF
--- a/src/swsssdk/port_util.py
+++ b/src/swsssdk/port_util.py
@@ -38,6 +38,9 @@ def get_interface_oid_map(db):
     """
     db.connect('COUNTERS_DB')
     if_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_PORT_NAME_MAP', blocking=True)
+    oid_pfx = len("oid:0x")
+    if_name_map = {if_name: sai_oid[oid_pfx:] for if_name, sai_oid in if_name_map.items()}
+
     if_id_map = {sai_oid: if_name for if_name, sai_oid in if_name_map.items()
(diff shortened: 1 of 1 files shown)


prsunny:
Changes to fix the side effects of Azure/sonic-sairedis#234

src/swsssdk/port_util.py
```diff
@@ -38,6 +38,9 @@ def get_interface_oid_map(db):
     """
     db.connect('COUNTERS_DB')
     if_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_PORT_NAME_MAP', blocking=True)
+    oid_pfx = len("oid:0x")
+    if_name_map = {if_name: sai_oid[oid_pfx:] for if_name, sai_oid in if_name_map.items()}
```
lguohan: sai_oid.split(':')[1]

prsunny: We would want to split till 0x.

qiluo-msft: >sai_oid [](start = 28, length = 7)

This change will bring dependency on schema version. Suggest to make it backward compatible, so we can still update snmp docker in old switches.

lguohan: there could be other changes in the counters_db or other new data in the database that is needed by new snmp. It will be very difficult to test all cases. Not sure we really need to have this as a goal.

PR #23, opened 2017-11-17
[configdb] fix bug of leaving unwanted columns in a hash

ishidawataru:
Current implementation doesn't remove unwanted columes with set_entry
like below.

```
>>> set_entry('table', 'key', {'key1': 'value1', 'key2': 'value2'})
>>> set_entry('table', 'key', {'key1': 'value3'})
>>> get_entry('table', 'key')
{'key1': 'value3', 'key2': 'value2'}
```

This commit fix this bug.

Signed-off-by: Wataru Ishida <ishida.wataru@lab.ntt.co.jp>

### DIFF
--- a/src/swsssdk/configdb.py
+++ b/src/swsssdk/configdb.py
@@ -155,7 +155,12 @@ class ConfigDBConnector(SonicV2Connector):
         if data == None:
             client.delete(_hash)
         else:
+            original = self.get_entry(table, key)
             client.hmset(_hash, self.__typed_to_raw(data))
+            for k in [ k for k in original.keys() if k not in data.keys() ]:
+                if type(original[k]) == list:
+                    k = k + '@'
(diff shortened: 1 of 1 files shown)


ishidawataru:
I found https://github.com/Azure/sonic-py-swsssdk/pull/15 solves the same issue in a different way.

https://github.com/Azure/sonic-py-swsssdk/pull/15 is more lightweight but needs to explicitly specify the keys for removal.
The fix I made doesn't need to specify the keys for removal but needs to retrieve the original data.

stcheng:
I don't think this is a bug. When we try to set an attribute with a new value, the current design is that we don't need to provide the rest of the attributes to replace the whole key-value tuple. The #15 will be enough for removing unwanted attributes.

lguohan:
set api usually does @ishidawataru says, we need to add a mod_entry API to explicit modify the field.

PR #25, opened 2017-12-07
[configdb]: add mod_entry, rename set_config to mod_config

lguohan:
mod_entry will not delete extra data in the db.

change the set_config to mod_config so that mod_config does not
delete extra data in the db

### DIFF
--- a/src/swsssdk/configdb.py
+++ b/src/swsssdk/configdb.py
@@ -222,7 +222,7 @@ class ConfigDBConnector(SonicV2Connector):
 
     def mod_config(self, data):
         """Write multiple tables into config db.
-           Extra entries/attributes in the db which are not in the data are kept.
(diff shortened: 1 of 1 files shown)


src/swsssdk/configdb.py
```diff
@@ -202,8 +220,8 @@ def get_table(self, table):
                 pass    #Ignore non table-formated redis entries
         return data
 
-    def set_config(self, data):
-        """Write multiple tables into config db. 
+    def mod_config(self, data):
+        """Write multiple tables into config db. Existing entries in the db are kept.
```
taoyl-ms: It's not "Existing entries are kept" (if existing entry occurs in `data` it will be modified), but "Untouched entries and untouched attributes of existing entries are kept".

lguohan: how about "extra entries/attributes in the db are kept"?

PR #26, opened 2017-12-13
[configdb]: Adopt get_table method to work with python3.

oleksandrivantsiv:
Decode data from redis DB into UTF-8 in python3.

### DIFF
--- a/src/swsssdk/configdb.py
+++ b/src/swsssdk/configdb.py
@@ -17,10 +17,11 @@ Example:
     config_db.listen()
 
 """
-
 import sys
 import time
 from .dbconnector import SonicV2Connector
+
+PY3K = sys.version_info >= (3, 0)
 
 class ConfigDBConnector(SonicV2Connector):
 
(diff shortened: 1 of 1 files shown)


src/swsssdk/configdb.py
```diff
@@ -17,11 +17,12 @@
     config_db.listen()
 
 """
-
 import sys
 import time
 from .dbconnector import SonicV2Connector
 
+PY3K = sys.version_info >= (3, 0)
```
qiluo-msft: >K [](start = 3, length = 1)

What does K mean?

src/swsssdk/configdb.py
```diff
@@ -101,16 +102,28 @@ def __raw_to_typed(self, raw_data):
         if raw_data == None:
             return None
         typed_data = {}
-        for key in raw_data:
+        for raw_key in raw_data:
+            key = raw_key
+            if PY3K:
```
qiluo-msft: >PY3K [](start = 15, length = 4)

Why we need to check python3? Is it possible not to?

oleksandrivantsiv: Library supposes to work with Python 2 and 3. Current implementation doesn't handle encoding correctly in Python 3. Python 2 has different API to works with encoding than Python 3. So it is not possible to not check.

src/swsssdk/configdb.py
lguohan: py3k is py3000, py3.0 same meaning.

jleveque: It appears "Python 3000" (or "py3k") was the initial name for what was eventually changed to "Python 3.0" to keep in line with the old versioning scheme before its actual release. I wasn't aware of the early name, and others might not be, either. We can always change it to "PY3X" or similar in the future.

Whatever we choose, we should probably use the same convention in all of our scripts whenever we need to make them compatible with both Python 2 and Python 3.

PR #34, opened 2018-08-29
support redis access setting via environment variables

ishidawataru:
Signed-off-by: Wataru Ishida <ishida@nel-america.com>

### DIFF
--- a/src/swsssdk/interface.py
+++ b/src/swsssdk/interface.py
@@ -1,5 +1,6 @@
 import time
 from functools import wraps
+import os
 
 import redis
 from redis import RedisError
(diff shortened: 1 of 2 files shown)


src/swsssdk/interface.py
```diff
@@ -149,9 +153,18 @@ def __init__(self, **kwargs):
...
+            if value:
+                kwargs[key[1]] = value
+
+        _set(self.REDIS_HOST)
+        _set(self.REDIS_PORT)
+        _set(self.REDIS_UNIX_SOCKET_PATH)
+
         self.redis_kwargs = kwargs
         if len(self.redis_kwargs) == 0:
-            self.redis_kwargs['unix_socket_path'] = self.REDIS_UNIX_SOCKET_PATH
```
qiluo-msft: >redis_kwargs [](start = 17, length = 12)

We always prefer unix socket than net socket because of performance. That's the reason we fallback to default REDIS_UNIX_SOCKET_PATH if nothing provided in kwargs.

qiluo-msft:
> class ConfigDBConnector(SonicV2Connector):

ConfigDBConnector is directly used by developer. Its construction parameter kwargs already supports
1. default unix socket path
2. custom unix socket path
3. custom net socket
I see the PR not necessary to provide the flexibility as the common library level. The design of these objects should similar to redis.StrictRedis [https://redis-py.readthedocs.io/en/latest/](https://redis-py.readthedocs.io/en/latest/)

---
Refers to: src/swsssdk/configdb.py:27 in 0fc8264. [](commit_id = 0fc826467872783eb0fc5c6e7a12f1a5496f4ffe, deletion_comment = False)

ishidawataru:
@qiluo-msft Thanks for your comment. I should have mentioned the motivation behind this change.

I'm deploying SONiC on top of ONL. In that environment, I can only access to redis via unix domain socket. However most of the tools from `sonic-utilities` is hard coded to use loopback interface instead of unix domain socket.

For example,

https://github.com/Azure/sonic-utilities/blob/master/scripts/portstat#L68

I opened this PR to tweak this behavior by using environment variables.
Maybe instead of this PR, I should open a PR for `sonic-utilites` to add an option to change the access method for redis?

ishidawataru:
@qiluo-msft Got it. I'll add a logic as you suggested and open a PR in `sonic-utilities` repo to support other methods to connect redis. Closing this.

PR #49, opened 2019-03-07
Enhance SDK to read from any DB by name

renukamanavalan:
Present code, hard codes the script to access Config-DB only. With this, this class can be used to read any DB. Added a get_keys method, so entire table need not be read into memory, but fetch values as and when needed.


renukamanavalan:
Tested the change using sonic-cfggen. Wrote another tool to compare APPL-DB & ASIC-DB  w.r.t routes and this tool uses this class to access either DB, which went successfully.

Care is taken to ensure that there is no change in pre-existing behavior.

PR #52, opened 2019-10-01
[multi-DB] Part 3: Python API changes

dzhangalibaba:
* add database_config.json parsing logic

* add  get_hostname/port/sokcet/dbid APIs

* make existing database.json contains all DB define, replace old database.json

* remove unnecessary check in dbconnector.py to count duplicated dbid values

* add SonicV2Connector connect() API to have the ability to choose db instances(host,port or socket) , then setting argvs and call DBInterface connector as what we did today.

* add a load interface to allow test cases to load config file.

* redesign DBInterface, SonicV2Connector
  -  DBinterface is the raw interface to connect a DB instance and accept the parameter as what we have today(hostname, port, socket) , the big changes is DBInterface APIs no longer consider DB name as parameter, it only accept raw dbid as parameter.

  - SonicDBConfig class handle all the db configuration related data. It store the mapping between dbname and dbid as well as the isntance info and socket/hostname/port onformation. SonicDBConfig also provide get API for others to access these data.

  - SonicV2Config class behaves as a bridge between SonicDBConfig and DBinterface. Its parent is DBinterface. Application use SonicV2Connector to connect to a db and write/read data from db. In SonicV2Connector, we do translation between dbname and dbid, also it provides the same signature as what the application is using today. Then we don't need to change the application codes.


src/swsssdk/dbconnector.py
```diff
@@ -17,23 +18,29 @@ def __init__(self, **kwargs):
 
 SonicV1Connector.db_map = _connector_map[SonicV1Connector.__name__]['db_map']
 
-if len(SonicV1Connector.db_map) != len({v['db'] for k, v in SonicV1Connector.db_map.items()}):
-    raise RuntimeError("Duplicate DB index detected in configuration.")
```
qiluo-msft: Why remove this block?

src/swsssdk/dbconnector.py
```diff
@@ -17,23 +18,29 @@ def __init__(self, **kwargs):
...
-
-
 class SonicV2Connector(DBInterface):
     def __init__(self, **kwargs):
         super(SonicV2Connector, self).__init__(**kwargs)
 
+    # after py test cases(e.g snmpagent/tests) are modified
+    # we can do the replacement on this new API via remnaming connect_new to connect
```
qiluo-msft: >remnaming [](start = 52, length = 9)

typo

src/swsssdk/dbconnector.py
```diff
@@ -17,23 +18,29 @@ def __init__(self, **kwargs):
...
+            self.redis_kwargs["unix_socket_path"] = None
+        super(SonicV2Connector, self).connect(db_name, retry_on)
+
     pass
 
 
 SonicV2Connector.db_map = _connector_map[SonicV2Connector.__name__]['db_map']
 
-if len(SonicV2Connector.db_map) != len({v['db'] for k, v in SonicV2Connector.db_map.items()}):
-    raise RuntimeError("Duplicate DB index detected in configuration.")
```
qiluo-msft: Why remove?

dzhangalibaba: > Why remove?

I added the missing LOGLEVEL_DB(3) , PFC_WD_DB(5), FLEX_COUNTERS_DB(5) into existing database.json which are mainly used for python today.  Since both PFC_WD_DB and FLEX_COUNTER_EB has value 5. This check will raise error. But these check is unnecessary, internally we are using string "DBNAME" as key to index(in SonicV2Connector/DBInterface), duplicated value index is fine,

src/swsssdk/dbconnector.py
dzhangalibaba: > Why remove this block?

I added the missing LOGLEVEL_DB(3) , PFC_WD_DB(5), FLEX_COUNTERS_DB(5) into existing database.json which are mainly used for python today.  Since both PFC_WD_DB and FLEX_COUNTER_EB has value 5. This check will raise error. But these check is unnecessary, internally we are using string "DBNAME" as key to index(in SonicV2Connector/DBInterface), duplicated value index is fine,
 
     "PFC_WD_DB": {
        "db": 5,
        "separator": ":"
      },
      "FLEX_COUNTER_DB": {
        "db": 5,
        "separator": ":"
      },

src/swsssdk/dbconnector.py
dzhangalibaba: fix typo

src/swsssdk/dbconnector.py
```diff
@@ -17,23 +18,29 @@ def __init__(self, **kwargs):
...
-
-
 class SonicV2Connector(DBInterface):
     def __init__(self, **kwargs):
         super(SonicV2Connector, self).__init__(**kwargs)
 
+    # after py test cases(e.g snmpagent/tests) are modified
+    # we can do the replacement on this new API via renaming connect_new to connect
+    # def connect(self, db_name, retry_on=True, isTcpconn=False):
+    def connect_new(self, db_name, retry_on=True, isTcpconn=False):
```
qiluo-msft: >connect_new [](start = 8, length = 11)

Is it possible to keep original function and change its implementation?

dzhangalibaba: The original connect() is from DBInterface class, which is not modified. Here I just overwrite the connect() in SonicV2Connector which is the child class to prepare some parameters, finally it is still call the original connect in DBInterface. now it is named as connect_new for temporally, since we didn't fix the py test cases yet which will cause test failed. Finally we will use name connect() which is the original name.

src/swsssdk/__init__.py
```diff
@@ -10,6 +10,53 @@
...
+    if _sonic_db_config_init == False:
+        _load_sonic_db_config()
+    inst_name = _sonic_db_config["DATABASES"][db_name]["instance"]
+    return _sonic_db_config["INSTANCES"][inst_name]["port"]
+
+def _get_sonic_db_id(db_name):
+    global _sonic_db_config_init
+    if _sonic_db_config_init == False:
+        _load_sonic_db_config()
+    return _sonic_db_config["DATABASES"][db_name]["id"]
```
qiluo-msft: Extract as a new class

src/swsssdk/dbconnector.py
dzhangalibaba: use connect() instead

src/swsssdk/__init__.py
dzhangalibaba: done

dzhangalibaba:
for the other installing a default database_config.json via lib topic we discussed offline, I tired today. Looks python wheel distribute cannot copy file to location out of the dist-package directory. The config file in whl must be in dist-package itself. Below is the link for this 'issue':
https://github.com/pypa/wheel/issues/92

so in the codes , I made some changes, if /var/run/redis/sonic-db/database_config.json is not there, we use the config at /usr/local/lib/python2.7/swsssdk/config/database_config.josn. This case only happened for py test cases, it is safe.

dzhangalibaba:
BTW VS test passed as well.

qiluo-msft:
>   }

The "db" fields are duplication with "id" fields in database_config.json. Some suggestions:

1. remove SonicV1Connector code, not used (better in a standalone PR)
2. remove this config file, and move separator info into database_config.json


---
Refers to: src/swsssdk/config/database.json:54 in a6aab64. [](commit_id = a6aab649dd7990af31a4b0b2317cc6bc034c5d4a, deletion_comment = False)

src/swsssdk/dbconnector.py
```diff
@@ -3,10 +3,61 @@
...
 
 # FIXME: Convert to metaclasses when Py2 support is removed. Metaclasses have unique interfaces to Python2/Python3.
 
+class SonicDBConfig(object):
+    SONIC_DB_CONFIG_FILE = "/var/run/redis/sonic-db/database_config.json"
+    _sonic_db_config_init = False
+    _sonic_db_config = {}
+
+    @staticmethod
+    def load_sonic_db_config(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
```
qiluo-msft: >load_sonic_db_config [](start = 8, length = 20)

I guess you could rewrite _load_connector_map instead of creating a new function.

src/swsssdk/dbconnector.py
```diff
@@ -3,10 +3,61 @@
...
+
+    @staticmethod
+    def get_sonic_db_port(db_name):
+        if SonicDBConfig._sonic_db_config_init == False:
+            SonicDBConfig.load_sonic_db_config()
+        inst_name = SonicDBConfig._sonic_db_config["DATABASES"][db_name]["instance"]
+        return SonicDBConfig._sonic_db_config["INSTANCES"][inst_name]["port"]
+
+    @staticmethod
+    def get_sonic_db_id(db_name):
```
qiluo-msft: >get_sonic_db_id [](start = 8, length = 15)

I guess you could rewrite get_dbid instead of creating a new function.

src/swsssdk/dbconnector.py
```diff
@@ -17,23 +68,28 @@ def __init__(self, **kwargs):
 
 SonicV1Connector.db_map = _connector_map[SonicV1Connector.__name__]['db_map']
 
-if len(SonicV1Connector.db_map) != len({v['db'] for k, v in SonicV1Connector.db_map.items()}):
-    raise RuntimeError("Duplicate DB index detected in configuration.")
-
-
 class SonicV2Connector(DBInterface):
     def __init__(self, **kwargs):
         super(SonicV2Connector, self).__init__(**kwargs)
```
qiluo-msft: >super(SonicV2Connector, self).__init__(**kwargs) [](start = 8, length = 48)

It makes no sense to reuse parent class `__init__`. Suggest add one option `is_unix_socket` for each object with default value true, and remove `isTcpconn` from `connect`

qiluo-msft:
Even with the difficulty in python wheel, we should try workaround to achieve the goal we discussed. Please help explorer. At least I know `python setup.py sdist` works, and of course this solution need some makefile modification.

---
In reply to: [538191085](https://github.com/Azure/sonic-py-swsssdk/pull/52#issuecomment-538191085) [](ancestors = 538191085)

dzhangalibaba:
Here having some root access issue as we discussed offline, I suggest keep the library  database_config.json file in the library domain , this won't affect anything on pyhsical machine. It helps solve the testing issue. For swss-common library, I suggest the same, with library there is a database_config.json file for testing only. Later when all the stuff are done, we can focus on how to solve the root access issue during building images and how to distribute the file to a single place. For now , to me it is OK, it doesn't affect any functionality .   

> @qiluo-msft
> for the other installing a default database_config.json via lib topic we discussed offline, I tired today. Looks python wheel distribute cannot copy file to location out of the dist-package directory. The config file in whl must be in dist-package itself. Below is the link for this 'issue':
> [pypa/wheel#92](https://github.com/pypa/wheel/issues/92)
> 
> so in the codes , I made some changes, if /var/run/redis/sonic-db/database_config.json is not there, we use the config at /usr/local/lib/python2.7/swsssdk/config/database_config.josn. This case only happened for py test cases, it is safe.

dzhangalibaba:
> > }
> 
> The "db" fields are duplication with "id" fields in database_config.json. Some suggestions:
> 
> 1. remove SonicV1Connector code, not used (better in a standalone PR)
> 2. remove this config file, and move separator info into database_config.json
> 
> Refers to: src/swsssdk/config/database.json:54 in [a6aab64](https://github.com/Azure/sonic-py-swsssdk/commit/a6aab649dd7990af31a4b0b2317cc6bc034c5d4a). [](commit_id = [a6aab64](https://github.com/Azure/sonic-py-swsssdk/commit/a6aab649dd7990af31a4b0b2317cc6bc034c5d4a), deletion_comment = False)

DONE

src/swsssdk/dbconnector.py
dzhangalibaba: done

src/swsssdk/dbconnector.py
dzhangalibaba: done

src/swsssdk/dbconnector.py
dzhangalibaba: done. invoking parent __init__ is necessary , since it initialized some variable/struct

dzhangalibaba:
@qiluo-msft redesign the codes, the changes is a little bit more, we can have a chat to explain what the idea is and how it works if possible.

src/swsssdk/dbconnector.py
```diff
@@ -1,20 +1,156 @@
...
+    def get_sonic_db_separator(db_name):
+        if SonicDBConfig._sonic_db_config_init == False:
+            SonicDBConfig.load_sonic_db_config()
+        return SonicDBConfig._sonic_db_config["DATABASES"][db_name]["separator"]
+
 class SonicV2Connector(DBInterface):
-    def __init__(self, **kwargs):
+    def __init__(self, unix_socket_path=True, **kwargs):
         super(SonicV2Connector, self).__init__(**kwargs)
+        self.use_unix_socket_path = unix_socket_path
```
qiluo-msft: >unix_socket_path [](start = 36, length = 16)

Suggestion: unix_socket_path->use_unix_socket_path

src/swsssdk/dbconnector.py
```diff
@@ -1,20 +1,156 @@
...
+        SonicDBConfig._sonic_db_config_init = True
+
+    @staticmethod
+    def get_sonic_db_list():
+        if SonicDBConfig._sonic_db_config_init == False:
+            SonicDBConfig.load_sonic_db_config()
+        return SonicDBConfig._sonic_db_config["DATABASES"].keys()
+
+    @staticmethod
+    def get_sonic_db_socket(db_name):
```
qiluo-msft: >get_sonic_db_socket [](start = 8, length = 19)

Suggest rename: get_sonic_db_* -> get_*

src/swsssdk/dbconnector.py
```diff
@@ -1,20 +1,156 @@
...
+        if SonicDBConfig._sonic_db_config_init == False:
+            SonicDBConfig.load_sonic_db_config()
+        inst_name = SonicDBConfig._sonic_db_config["DATABASES"][db_name]["instance"]
```
qiluo-msft: >db_name [](start = 64, length = 7)

Should we handle index out of range explicitly?
1. if db_name does not exist, it's invalid argument
2. if inst_name does not exist, it's invalid config.

qiluo-msft: Extract a common helper function like get_instance?

src/swsssdk/dbconnector.py
```diff
@@ -1,20 +1,156 @@
...
+    def load_sonic_db_config(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
+        """
+        Get multiple database config from the database_config.json
+        """
+        if SonicDBConfig._sonic_db_config_init == True:
+            return
+
+        try:
+            if os.path.isfile(sonic_db_file_path) == False:
+                sonic_db_file_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'config', 'database_config.json')
```
qiluo-msft: >sonic_db_file_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'config', 'database_config.json') [](start = 16, length = 111)

We don't want to see this case in production. Can we add a syslog warning?

src/swsssdk/dbconnector.py
dzhangalibaba: DONE

src/swsssdk/dbconnector.py
dzhangalibaba: DONE

src/swsssdk/dbconnector.py
dzhangalibaba: added db_name_validation() and inst_name_validation(), if invalid , raise a error

dzhangalibaba: DONE

src/swsssdk/dbconnector.py
dzhangalibaba: added warning log

dzhangalibaba:
> Thanks! Minor issues remain.

updated

PR #59, opened 2020-01-14
[MultiDB]: update sonic-db-cli output to redis-cli output format

dzhangalibaba:
* When field value is empty:
    * today redis-cli return "", sonic-db-cli return 'None'
    * some caller check 'None' word since the data may be string 'None' 
    * CHANGES, sonic-db-cli check return type , if it is None, then return "", if not None, return the string (including "None")

* when return value is a list:
    * today redis-cli returns "TB|E1 TB|E2 TB|E3 ..." , sonic-db-cli returns "['TB|E1',  'TB|E2',  'TB|E3'] ..", when list is empty, redis-cli return "" while sonic-db-cli returns [].
    * CHANGES, sonic-db-cli check return type, it it is a list, output all thge element as a joint string.

* So far, only found these two usage is needed to be changed in sonic-db-cli , other redis operation like SET/FLUSH/GET/HGET/KEYS are good and the output are some format after these changes.

### DIFF
--- a/src/swsssdk/scripts/sonic-db-cli
+++ b/src/swsssdk/scripts/sonic-db-cli
@@ -25,9 +25,14 @@ else:
         client = dbconn.get_redis_client(dbname)
         args = sys.argv[2:]
         resp = client.execute_command(*args)
+        """
+        sonic-db-cli output format mimic the non-tty mode output format from redis-cli
+        based on our usage in SONiC, None and list type output from python API needs to be modified
+        with these changes, it is enough for us to mimic redis-cli in SONiC so far since no application uses tty mode redis-cli output
+        """
         if resp is None:
             print ""
-        elif type(resp) is list:
+        elif isinstance(resp, list):
             print " ".join(resp)
         else:
             print resp
--- a/src/swsssdk/scripts/sonic-db-cli
+++ b/src/swsssdk/scripts/sonic-db-cli
@@ -24,4 +24,10 @@ else:
     else:
         client = dbconn.get_redis_client(dbname)
         args = sys.argv[2:]
-        print client.execute_command(*args)
+        resp = client.execute_command(*args)
+        if resp is None:
+            print ""
+        elif type(resp) is list:
+            print " ".join(resp)
+        else:
+            print resp


src/swsssdk/scripts/sonic-db-cli
```diff
@@ -24,4 +24,10 @@ else:
     else:
         client = dbconn.get_redis_client(dbname)
         args = sys.argv[2:]
-        print client.execute_command(*args)
+        resp = client.execute_command(*args)
+        if resp is None:
+            print ""
+        elif type(resp) is list:
```
qiluo-msft: >type(resp) is list [](start = 13, length = 18)

isinstance(resp, list)

qiluo-msft:
> #!/usr/bin/python

Could you add code comment that the behavior is mimic redis-cli non-tty mode, which is different from redis-cli tty mode?

---
Refers to: src/swsssdk/scripts/sonic-db-cli:1 in 17bb905. [](commit_id = 17bb90566b1fbf33bf47fd80659a5f117ad38569, deletion_comment = False)

src/swsssdk/scripts/sonic-db-cli
dzhangalibaba: DONE

dzhangalibaba:
> > #!/usr/bin/python
> 
> Could you add code comment that the behavior is mimic redis-cli non-tty mode, which is different from redis-cli tty mode?
> 
> Refers to: src/swsssdk/scripts/sonic-db-cli:1 in [17bb905](https://github.com/Azure/sonic-py-swsssdk/commit/17bb90566b1fbf33bf47fd80659a5f117ad38569). [](commit_id = [17bb905](https://github.com/Azure/sonic-py-swsssdk/commit/17bb90566b1fbf33bf47fd80659a5f117ad38569), deletion_comment = False)

ADDED

qiluo-msft:
@dzhangalibaba Please add some unit test for sonic-db-cli.

dzhangalibaba:
> @dzhangalibaba Please add some unit test for sonic-db-cli.

OK. I'll add some test cases into vs tests later.

qiluo-msft:
> > @dzhangalibaba Please add some unit test for sonic-db-cli.
> 
> OK. I'll add some test cases into vs tests later.

To test this only, you'd better use googletest, like sonic-swss-common.

dzhangalibaba:
> > > @dzhangalibaba Please add some unit test for sonic-db-cli.
> > 
> > 
> > OK. I'll add some test cases into vs tests later.
> 
> To test this only, you'd better use googletest, like sonic-swss-common.

That will be a new project to add something like that. Currently,  pyswsssdk is not tested on real redis server, we only have some snmp py related unit tests based on the mock tables when building the .whl pkg in built environment. We don't have test framework which already set up redis server for testing like vs tests and sairedis tests did..  I suggest to investigate/discuss how to add py library related UT on real redis server in parallel, then we can find the best solution, this may be time consuming. To unblock the other work, I suggest go ahead with current PR and continue the sonic-db-cli replacement activities. These py lib related UT addition start in parallel.

dzhangalibaba:
@qiluo-msft can we merge this first , then I can sent out three other pending PRs depends on this.

PR #63, opened 2020-03-05
Namespace support for the python classes.

judyjoseph:
- What I did

This PR is the changes needed to support multiple namespaces for the Multi-ASIC devices. Multi-DB namespace PR --> (https://github.com/Azure/SONiC/pull/567)

- How I did it

The changes are mainly to these classes 
1. SonicDBConfig
2. SonicV2Connector/ConfigDBConnector

**A new parameter "namespace" is added to the SonicV2Connector class __init__ , to pass the namespace name. The default value of <namespace> is kept as '' which is an empty string representing that user didn't give a namespace name.**

**class SonicV2Connector(DBInterface):**
   def __init__(self, use_unix_socket_path=False, **namespace=''**, **kwargs):

If the user don't explicitly set this parameter, namespace takes '' as value, and connects to the db_name in the local context. When I mean local context, it refers to the database docker running in the current namespace wherever  you are running this application/script …. ( It could be either Linux host or any specific network namespace ). In this way it is backward compatible and the existing behavior of APIs in these utility classes are maintained.

In the SonicDBConfig, a new API is introduced load_sonic_global_db_config() which loads the /var/run/redis/sonic-db/database_global.json file if present. This file has the mapping between the namespace name and the corresponding database_config.json file.

- How to verify it
Verified the dockers come up in the Multi-ASIC platform and the config commands too were verified to an extend.

- Description for the changelog

- A picture of a cute animal (not mandatory but encouraged)


qiluo-msft:
> import swsssdk

Suggest using argparse instead of reinvent the wheel.

---
Refers to: src/swsssdk/scripts/sonic-db-cli:3 in e61c5e6. [](commit_id = e61c5e6f6fd919bff3f3de9659b2fd83d95f43a4, deletion_comment = False)

src/swsssdk/dbconnector.py
```diff
@@ -26,165 +29,384 @@ def load_sonic_db_config(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
...
+
+    """
+    Connect to the DB instance specified by the namespace parameter in database_config.json file.
+    if namespace="default", it connects to the default instance of "db_name".
+    if namespace=<0...n>, any value between 0 and n,
+                          it connects to "db_name" instance in that namespace as per the INCLUDES.
+    if namespace="all", it connects to the global instance of "db_name" + namespace <0..n> as present in INCLUDES.
+    """
+    def connect(self, db_name, retry_on=True, namespace='default'):
+        self.ns_name = namespace
```
qiluo-msft: >self.ns_name = namespace [](start = 8, length = 24)

we should throw NotImplmentedError when namespace is used with TCP port.

src/swsssdk/dbconnector.py
```diff
@@ -26,165 +29,384 @@ def load_sonic_db_config(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
...
-    def delete(self, db_name, key, *args, **kwargs):
-        db_id = self.get_dbid(db_name)
-        return super(SonicV2Connector, self).delete(db_id, key, *args, **kwargs)
+    def keys(self, db_name, pattern='*', namespace='none', *args, **kwargs):
+        if namespace == 'none':
+            namespace = self.ns_name
+
+        data = []
+        ns_instances = []
+        if namespace == 'all':
```
qiluo-msft: >if namespace == 'all': [](start = 8, length = 22)

This 'all' namespace seems not fundamental feature of SonicV2Connector class. Could you extract this feature outside this class?

src/swsssdk/dbconnector.py
```diff
@@ -26,165 +29,384 @@ def load_sonic_db_config(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
...
+            return get_keys_and_all(db_name, _hash, namespace, *args, **kwargs)
+
+        if namespace == 'none':
+            namespace = self.ns_name
+
+        db_id = self.get_dbid(db_name, namespace)
+        return self.db_intfs[namespace].get_all(db_id, _hash, *args, **kwargs)
+
+    def set(self, db_name, _hash, key, val, namespace='none', *args, **kwargs):
+        if namespace == 'none':
```
qiluo-msft: >namespace == 'none' [](start = 11, length = 19)

`none` is a valid namespace name. you may use `if namespace:`

judyjoseph:
> import swsssdk
> 
> Suggest using argparse instead of reinvent the wheel.
> 
> Refers to: src/swsssdk/scripts/sonic-db-cli:3 in e61c5e6. [](commit_id = e61c5e6, deletion_comment = False)

# Updated the sonic-db-cli implementation using argparse.

src/swsssdk/dbconnector.py
judyjoseph: Done, NotImplementedError is thrown if the user attempts to use TCP socket connection and to a db instance not in the 'default' redis instance.

src/swsssdk/dbconnector.py
judyjoseph: Understand your comment that "all", "none" etc could be valid namespace names. 

I have made a few changes to the options, as below to connect to the databases. Introduced a new parameter "mode"

"namespace"
	'default' : it connects to the local 'db_name' instance.
	<asic0,asic1 ...asicn> : it connects to "db_name" instance in that namespace
"mode"
	'exclusive' : connect exclusively to the db instance in the namespace specified.
	'all' : connect to the db instance of the "db_name" in all namespaces.

I would like to have the SonicV2 Connector class and ConfigDBConnector class able to connect to a specific db_instance exclusively -- OR -- connect to the db_instances in all namespaces.

I suggest to keep it this way as we have seen many use cases where we need to get/set tables in a similar fashion on the db_instances in all namespaces. With the utility class supporting 'all' we can avoid adding for loops every where in the code ... eg: in get/get_all in case of SNMP, Config CLI start/shut all BGP neighbors etc

If user don't specifiy these parameters, ( namespace, mode ) will take ('default' , 'exclusive') which connects to the specific db_instance in the local redis server. This helps to retain the current behavior of these utility classes, if the user don't pass any arguments.

src/swsssdk/dbconnector.py
judyjoseph: Updated the code and used None keyword instead.

src/swsssdk/scripts/sonic-db-cli
```diff
@@ -1,38 +1,51 @@
...
+      Example 2: sonic-db-cli -ns asic1 APPL_DB HGETALL VLAN_TABLE:Vlan10
+      Example 3: sonic-db-cli APPL_DB HGET VLAN_TABLE:Vlan10 mtu
+      Example 4: sonic-db-cli -ns asic5 APPL_DB EVAL "return {KEYS[1],KEYS[2],ARGV[1],ARGV[2]}" 2 k1 k2 v1 v2
+'''
+
+def main():
+    parser = argparse.ArgumentParser(description='SONiC DB CLI',
+                         version='1.0.0',
+                         formatter_class=argparse.RawTextHelpFormatter)
+    parser.add_argument('-ns', '--namespace', type=str, help='Namespace to use [asic0,asic1...,asicn]', default='default')
```
qiluo-msft: >-ns [](start = 25, length = 3)

prefer `-n`

src/swsssdk/scripts/sonic-db-cli
```diff
@@ -1,38 +1,51 @@
...
+    parser.add_argument('database', nargs='?', type=str, help='Database name')
```
qiluo-msft: >database [](start = 25, length = 8)

add `required`, so no need to check yourself.

src/swsssdk/scripts/sonic-db-cli
```diff
@@ -1,38 +1,51 @@
...
+    parser.add_argument('cmd', nargs='+', type=str, help='Command to execute in database')
```
qiluo-msft: >cmd [](start = 25, length = 3)

the same

src/swsssdk/scripts/sonic-db-cli
```diff
@@ -1,38 +1,51 @@
...
+            """
+            if resp is None:
+                print ""
+            elif isinstance(resp, list):
+                print "\n".join(resp)
+            else:
+                print resp
+    else:
+        parser.print_help()
+        sys.exit(1)
```
qiluo-msft: Add new testcase for new code

src/swsssdk/configdb.py
```diff
@@ -170,37 +220,70 @@ def set_entry(self, table, key, data):
...
+                client.delete(_hash)
+            else:
+                original = self.get_entry(table, key)
+                client.hmset(_hash, self.__typed_to_raw(data))
+                for k in [ k for k in original.keys() if k not in data.keys() ]:
+                    if type(original[k]) == list:
+                        k = k + '@'
+                    client.hdel(_hash, self.serialize_key(k, instance))
+
+    def mod_entry(self, table, key, data, create_new_flag=False, namespace=None):
```
qiluo-msft: >mod_entry [](start = 8, length = 9)

I see reason to add `namespace` to `connect()`, but no reason to add each other function.

src/swsssdk/dbconnector.py
judyjoseph: I thought about the various use cases and came up with this new approach for user to pass a list of namespaces who he wants to connect to. I thought about having separate classes for single namespace and multi namespace -- but found that there is so much of code duplication if I do so.

Additionally removing the namespace parameter from the APIs helped reduce the new code added.

src/swsssdk/dbconnector.py
```diff
@@ -9,182 +9,324 @@
...
+            if os.path.isfile(sonic_db_file_path) ==  True:
+                with open(sonic_db_file_path, "r") as read_file:
+                    all_dbs = json.load(read_file)
+                    for entry in all_dbs['INCLUDES']:
+                        if 'namespace' not in entry.keys():
+                            namespace = 'default'
+                        else: 
+                            namespace = entry['namespace']
+
+                        db_include_file = SonicDBConfig.SONIC_DB_BASE_DIR+entry['include']
```
qiluo-msft: >+ [](start = 73, length = 1)

Use `os.path.join()` instead of `+`

qiluo-msft: >entry['include'] [](start = 74, length = 16)

We need to support `include` relative path relative to `sonic_db_file_path`. Not relative to SONIC_DB_BASE_DIR.

src/swsssdk/dbconnector.py
```diff
@@ -9,182 +9,324 @@
...
     _sonic_db_config = {}
 
     @staticmethod
-    def load_sonic_db_config(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
+    def load_sonic_db_config(sonic_db_file_path=SONIC_DB_GLOBAL_CONFIG_FILE):
         """
-        Get multiple database config from the database_config.json
+        Parse and load the database config json files
+        Check if either of SONIC_DB_GLOBAL_CONFIG_FILE or SONIC_DB_CONFIG_FILE is present, unless the 
+        'sonic_db_file_path' is explicitly passed as parameter.
```
qiluo-msft: Function signature is not well designed. Considering below cases:

1. keep backward-compatibility: use `SONIC_DB_CONFIG_FILE` if this function is not called
2. keep backward-compatibility: use `sonic_db_file_path` if this function is not called
3. add new feature (maybe a new function) to support load database_global.json user explicitly specify
3. I *think* there is no need to have a default SONIC_DB_GLOBAL_CONFIG_FILE. This is new requirement and usage is quite limited.

src/swsssdk/dbconnector.py
```diff
@@ -9,182 +9,324 @@
...
+
         SonicDBConfig._sonic_db_config_init = True
 
     @staticmethod
-    def db_name_validation(db_name):
+    def db_name_validation(db_name, ns_instance):
         if SonicDBConfig._sonic_db_config_init == False:
             SonicDBConfig.load_sonic_db_config()
-        if db_name not in SonicDBConfig._sonic_db_config["DATABASES"]:
+        data=SonicDBConfig._sonic_db_config[ns_instance]["DATABASES"]
```
qiluo-msft: >data [](start = 8, length = 4)

db or databases

src/swsssdk/dbconnector.py
```diff
@@ -9,182 +9,324 @@
...
         except (OSError, IOError):
             msg = "Could not open sonic database config file '{}'".format(sonic_db_file_path)
             logger.exception(msg)
             raise RuntimeError(msg)
+
         SonicDBConfig._sonic_db_config_init = True
 
     @staticmethod
-    def db_name_validation(db_name):
+    def db_name_validation(db_name, ns_instance):
```
qiluo-msft: >ns_instance [](start = 36, length = 11)

just call it ns or namespace

src/swsssdk/dbconnector.py
```diff
@@ -9,182 +9,324 @@
...
+                            SonicDBConfig._sonic_db_config[namespace] = json.load(inc_file)
+            else:
+                sonic_db_file_path = SonicDBConfig.SONIC_DB_CONFIG_FILE
+                if os.path.isfile(sonic_db_file_path) ==  False:
+                    msg = "'{}' is not found, it is not expected in production devices!!".format(sonic_db_file_path)
+                    logger.warning(msg)
+                    sonic_db_file_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'config', 'database_config.json')
+
+                with open(sonic_db_file_path, "r") as read_file:
+                    SonicDBConfig._sonic_db_config['default'] = json.load(read_file)
```
qiluo-msft: >'default' [](start = 51, length = 9)

`default` is a valid namespace name. let's use `''` as your default.

src/swsssdk/dbconnector.py
qiluo-msft: >ns_instance [](start = 44, length = 11)

what happens if ns_instance does not exist in _sonic_db_config?

src/swsssdk/dbconnector.py
```diff
@@ -9,182 +9,324 @@
...
 
     @staticmethod
-    def get_dblist():
+    def get_dblist(ns_instance):
+        if SonicDBConfig._sonic_db_config_init == False:
+            SonicDBConfig.load_sonic_db_config()
+        return SonicDBConfig._sonic_db_config[ns_instance]["DATABASES"].keys()
+
+    @staticmethod
+    def is_instance_valid(ns_instance):
```
qiluo-msft: >is_instance_valid [](start = 8, length = 17)

`instance` is ambigous

qiluo-msft:
@dzhangalibaba, could you help review?

src/swsssdk/dbconnector.py
```diff
@@ -9,182 +9,324 @@
...
+            self.db_connect(db_name, retry_on, instance)
 
     def close(self, db_name):
-        db_id = self.get_dbid(db_name)
-        super(SonicV2Connector, self).close(db_id)
+        for instance in ns_instances:
+            db_id = self.get_dbid(db_name, instance)
+            self.db_intfs[instance].close(db_id)
+
+    def check_instance_valid(self, namespace='default'):
```
qiluo-msft: >check_instance_valid [](start = 8, length = 20)

Why this is needed? Give a sample use case?

src/swsssdk/dbconnector.py
```diff
@@ -9,182 +9,324 @@
 # FIXME: Convert to metaclasses when Py2 support is removed. Metaclasses have unique interfaces to Python2/Python3.
 
 class SonicDBConfig(object):
```
qiluo-msft: >SonicDBConfig [](start = 6, length = 13)

You may define a new class (SonicNamespaceConfig) above original SonicDBConfig. The original class fits well in this layered design.

src/swsssdk/dbconnector.py
```diff
@@ -9,182 +9,324 @@
...
             SonicDBConfig.load_sonic_db_config()
-        return SonicDBConfig._sonic_db_config["DATABASES"].keys()
+        if SonicDBConfig._sonic_db_config[ns_instance] != None:
+            return True
+        else: 
+            return False
 
     @staticmethod
-    def get_instance(db_name):
+    def get_dbConnection_instances():
```
qiluo-msft: >get_dbConnection_instances [](start = 8, length = 26)

get_namespace_list?

src/swsssdk/dbconnector.py
```diff
@@ -9,182 +9,324 @@
...
+        return got_data
 
     def get_all(self, db_name, _hash, *args, **kwargs):
-        db_id = self.get_dbid(db_name)
-        return super(SonicV2Connector, self).get_all(db_id, _hash, *args, **kwargs)
+        # If there are more than one namespace call get_keys_and_all
+        if len(ns_instances) > 1:
+            return get_keys_and_all(db_name, _hash, *args, **kwargs)
+
+        db_id = self.get_dbid(db_name, self.ns_name)
```
judyjoseph: self.ns_name, needs to be changed.

src/swsssdk/scripts/sonic-db-cli
judyjoseph: Since these are positional arguments, cannot add the required option.

src/swsssdk/scripts/sonic-db-cli
judyjoseph: Since these are positional arguments, cannot add the required option. Hence keeping the same implementation.

src/swsssdk/dbconnector.py
judyjoseph: Updated.

src/swsssdk/dbconnector.py
judyjoseph: Changed the logic to keep the old signature of taking SONIC_DB_CONFIG_FILE as the defaut incase the user hasn't specified the file. 

The additional logic needed now is to check if SONIC_DB_GLOBAL_CONFIG_FILE is present in DIR derived from sonic_db_file_path. 
If present --> parse the SONIC_DB_GLOBAL_CONFIG_FILE 
if not present --> continue parsing the file at sonic_db_file_path

Let me know your comments on this approach.

src/swsssdk/dbconnector.py
judyjoseph: Updated.

src/swsssdk/dbconnector.py
judyjoseph: Updated.

src/swsssdk/dbconnector.py
judyjoseph: Changed to using '' instead of 'default'

src/swsssdk/dbconnector.py
judyjoseph: Updated.

src/swsssdk/dbconnector.py
judyjoseph: Updated by changing instance to namespace at most places.

src/swsssdk/dbconnector.py
judyjoseph: Done

src/swsssdk/dbconnector.py
judyjoseph: I have added a new API namespace_validation() introduced to check the namespace validation similar to instance/db_name.

src/swsssdk/dbconnector.py
judyjoseph: The idea was to call this API, when the user passes the namespace as an argument. We take the argument he passed, and check if it is there in the list of namespaces which we get by parsing the database_config.json files. 

I have updated the usage now with a similar API namespace_validation() introduced to check the namespace validation similar to instance/db_name. This is better than doing this check at the connect() APIs.

src/swsssdk/dbconnector.py
judyjoseph: I find from a bit of testing that it is easier to extract/keep the namespace details in SonicDBConfig itself, since we have the "namespace information" also in the database json files which we have to parse in the SonicDBConfig class.

Additionally this will help keep the entry point same as SonicDBConfig class irrespective of whether the SonicV2Connector/ConfigDBConnector is used with a "specific namespace" or not.

src/swsssdk/dbconnector.py
```diff
@@ -9,182 +9,324 @@
...
+    """
+    Connect to the DB instance in the "namespaces" specified. It is a list.
+	* [''] : It connects to the local 'db_name' instance. This is the default argument.
+                 So if the user needs to talk to the local DB in a particular namespace
+                 call connect() API without the namespaces argument.
+        * list of namespaces (1 or more ) to connect to ['asic0', 'asic1', 'asic2' ...]
+    """
+    def connect(self, db_name, retry_on=True, namespaces=['']):
+        # The TCP connection to a DB in different namespace in not supported.
+        if namespaces != [''] and self.use_unix_socket_path == False:
```
qiluo-msft: >namespaces [](start = 11, length = 10)

What if `namespaces is None`? treat it as same as `['']` or invalid argument?

src/swsssdk/dbconnector.py
```diff
@@ -9,182 +9,324 @@
...
             raise RuntimeError(msg)
+
         SonicDBConfig._sonic_db_config_init = True
 
     @staticmethod
-    def db_name_validation(db_name):
+    def namespace_validation(namespace):
+        if SonicDBConfig._sonic_db_config_init == False:
+            SonicDBConfig.load_sonic_db_config()
+        if namespace not in SonicDBConfig._sonic_db_config.keys():
```
qiluo-msft: >SonicDBConfig._sonic_db_config.keys() [](start = 28, length = 37)

You can remove `.keys()`

src/swsssdk/dbconnector.py
```diff
@@ -9,182 +9,324 @@
...
 
     @staticmethod
-    def inst_name_validation(inst_name):
+    def inst_name_validation(inst_name, namespace):
         if SonicDBConfig._sonic_db_config_init == False:
             SonicDBConfig.load_sonic_db_config()
-        if inst_name not in SonicDBConfig._sonic_db_config["INSTANCES"]:
+        SonicDBConfig.namespace_validation(namespace)
+        instances = SonicDBConfig._sonic_db_config[namespace]["INSTANCES"]
+        if inst_name not in instances.keys():
```
qiluo-msft: >.keys() [](start = 37, length = 7)

You can remove `.keys()`

src/swsssdk/dbconnector.py
```diff
@@ -9,182 +9,324 @@
...
             msg = "{} is not a valid instance name in configuration file".format(inst_name)
             logger.exception(msg)
```
qiluo-msft: >logger.exception [](start = 12, length = 16)

This function should only be called from an exception handler

src/swsssdk/dbconnector.py
```diff
@@ -9,182 +9,324 @@
...
+        if os.path.isfile(global_db_file_path) ==  True:
+            with open(global_db_file_path, "r") as read_file:
+                all_dbs = json.load(read_file)
+                for entry in all_dbs['INCLUDES']:
+                    if 'namespace' not in entry.keys():
+                        ns = ''
+                    else: 
+                        ns = entry['namespace']
+
+                    db_include_file = os.path.join(SonicDBConfig._sonic_db_config_dir, entry['include'])
```
qiluo-msft: >db_include_file = os.path.join(SonicDBConfig._sonic_db_config_dir, entry['include']) [](start = 20, length = 84)

We need to support `include` relative path relative to `global_db_file_path`. Not relative to SONIC_DB_BASE_DIR.

src/swsssdk/scripts/sonic-db-cli
judyjoseph: Updated test/test_moduleLoad.py for handling the database_global.json file and namespace changes to SonicDBConfig class. 

I have verified the sonic-db-cli by making sure the docker services are up on a DUT with  these changes. If we need to add specific py testcases, we would need to use something like mock redis client.

src/swsssdk/dbconnector.py
judyjoseph: I have now added a new API **load_sonic_db_configs()** to explicitly invoke load_sonic_global_db_config() and load_sonic_db_config(). This API **load_sonic_db_configs()** needs to be invoked from the SonicV2Connector class constructor.

In the new API load_sonic_global_db_config() if the file /var/run/redis/sonic-db/database_global.json, is present, it will be parsed and loaded.

This way I am **not** modifying the behavior of existing API load_sonic_db_config().

src/swsssdk/dbconnector.py
```diff
@@ -27,90 +67,134 @@ def load_sonic_db_config(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
...
+    @staticmethod
+    def get_separator(db_name, namespace=''):
+        if SonicDBConfig._sonic_db_config_init == False:
+            SonicDBConfig.load_sonic_db_config()
+        SonicDBConfig.db_name_validation(db_name, namespace)
+        return SonicDBConfig._sonic_db_config[namespace]["DATABASES"][db_name]["separator"]
 
 class SonicV2Connector(DBInterface):
-    def __init__(self, use_unix_socket_path=False, **kwargs):
+    def __init__(self, use_unix_socket_path=True, **kwargs):
```
qiluo-msft, line 170: >True [](start = 44, length = 4)

Change default behavior will break backward compatibility.

src/swsssdk/dbconnector.py
```diff
@@ -27,90 +67,134 @@ def load_sonic_db_config(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
...
         super(SonicV2Connector, self).__init__(**kwargs)
         self.use_unix_socket_path = use_unix_socket_path
+        self.namespace = ''
+
+        # Load sonic DB configs
+        self.load_db_configs()
```
qiluo-msft, line 176: This will change old behavior.

src/swsssdk/dbconnector.py
```diff
@@ -27,90 +67,134 @@ def load_sonic_db_config(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
...
+
         for db_name in self.get_db_list():
             # set a database name as a constant value attribute.
             setattr(self, db_name, db_name)
 
-    def connect(self, db_name, retry_on=True):
+    def connect(self, db_name, retry_on=True, namespace=''):
+        if namespace == None:
```
qiluo-msft, line 183: >if namespace == None: [](start = 8, length = 21)

Sorry for change mind on old comment, I believe `isinstance(namespace, str)` will be even better.

src/swsssdk/dbconnector.py
```diff
@@ -27,90 +67,134 @@ def load_sonic_db_config(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
...
     @staticmethod
-    def db_name_validation(db_name):
+    def load_sonic_db_configs():
+        if SonicDBConfig._sonic_db_global_config_init == False:
+            SonicDBConfig.load_sonic_global_db_config()
+        if SonicDBConfig._sonic_db_config_init == False:
+            SonicDBConfig.load_sonic_db_config()
+
+    @staticmethod
+    def namespace_validation(namespace=''):
```
qiluo-msft, line 85: >='' [](start = 38, length = 3)

I believe required parameter make more sense.

src/swsssdk/dbconnector.py
```diff
@@ -27,90 +67,134 @@ def load_sonic_db_config(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
...
+
+        # The TCP connection to a DB in different namespace in not supported.
+        if namespace != '' and self.use_unix_socket_path == False:
+            message = "TCP connectivity to the DB instance in a different namespace is not implemented!"
+            raise NotImplementedError(message)
+
+        self.namespace = namespace
+        for dbname in self.get_db_list():
+            # set a database name as a constant value attribute.
+            setattr(self, dbname, dbname)
```
qiluo-msft, line 196: Why add this part? It is also inefficient to execute for each `connect()`

src/swsssdk/dbconnector.py
judyjoseph: The default behavior was originally use_unix_socket_path=True, but was reverted in this commit. Do you remember any specific reason Qi ?

commit bc3964b788c3a4a45f2b359a5df5934ecdee84c2
Author: Dong Zhang <41927498+dzhangalibaba@users.noreply.github.com>
Date:   Tue Nov 5 09:13:05 2019 -0800

I remember originally unix socket was used as default as it was faster than TCP socket.

src/swsssdk/dbconnector.py
judyjoseph: I need this new behaviour where I have to load the **database_global.json** file in addition to the database_config.json file. If we are in the linux host both these files will be present in the "/var/run/redis/sonic-db". 

If we are in a particular namespace ( for eg: namespace 1) we will only have the file database_config.json alone in the directory  /var/run/**redis1**/sonic-db/" which is mapped to --> "/var/run/redis/sonic-db" in the namespace.

So this API load_db_configs() .. will load either database_config.json alone .. or will load both database_global.json file and database_config.json if present.

src/swsssdk/dbconnector.py
judyjoseph: I have to do it per connect, where the user specifies the namespace and the databases could be different as defined in the database_config.json ( present in that namespace ). I think it is better to assume that the dbnames could be different in different namespaces ?

I could probably add a check to see if the namespace user given is the same as default namespace '' -- in which case I will ignore doing setattr again ( as for default namespace '' it is already done in __init__ constructor of SonicV2Connector class )

qiluo-msft: I agree the dbnames could be different in different namespaces. What is the use case of these constant value attributes? I cannot find one in this PR.

---
In reply to: [402695930](https://github.com/Azure/sonic-py-swsssdk/pull/63#discussion_r402695930) [](ancestors = 402695930)

src/swsssdk/dbconnector.py
qiluo-msft: The default behavior of original code should be kept if possible. I understand your 'new behavior', and you could explicitly invoke SonicDBConfig functions in your limited new use cases (new applications) to achieve the 'new behavior'. But do not change old behavior.

---
In reply to: [402693884](https://github.com/Azure/sonic-py-swsssdk/pull/63#discussion_r402693884) [](ancestors = 402693884)

src/swsssdk/dbconnector.py
```diff
@@ -27,90 +67,134 @@ def load_sonic_db_config(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
...
+
+        # Load sonic DB configs
+        self.load_db_configs()
+
         for db_name in self.get_db_list():
             # set a database name as a constant value attribute.
             setattr(self, db_name, db_name)
 
-    def connect(self, db_name, retry_on=True):
+    def connect(self, db_name, retry_on=True, namespace=''):
```
qiluo-msft, line 182: >, namespace='' [](start = 44, length = 14)

This is override function and supposed to keep the function signature. How about move namespace parameter to `__init__()` and limit one instance could only connect to databases in one namespace.

src/swsssdk/dbconnector.py
judyjoseph: I did try loading the image in the DUT and it was coming up fine.
But since you are echoing issues you faced earlier setting default as UNIX_socket , I would revert back to default use_unix_socket_path=False.

src/swsssdk/dbconnector.py
judyjoseph: Updated to handle the LGTM code analysis tool above .. so it is taken care.

src/swsssdk/dbconnector.py
judyjoseph: Ok Qi, Agreed. I have changed the logic so that we don't explicitly load the db_configs() as before. 

I made it as a part of the namespace validation logic. In this below API, if the namespace is not equal to default '' namespace, this means a user has explicitly asked for an external namespace. In this case we load the global config if it is not loaded.

    @staticmethod
    def namespace_validation(namespace):
        # Load global config if the namespace is an external one.
        if namespace != '' and SonicDBConfig._sonic_db_global_config_init == False:
            SonicDBConfig.load_sonic_global_db_config()

This way I believe I am not touching the old behavior .. as in the old behavior namespace argument defaults to ''

qiluo-msft:
@dzhangalibaba, could you help review?

src/swsssdk/scripts/sonic-db-cli
```diff
@@ -1,38 +1,52 @@
...
-        sonic-db-cli output format mimic the non-tty mode output format from redis-cli
-        based on our usage in SONiC, None and list type output from python API needs to be modified
-        with these changes, it is enough for us to mimic redis-cli in SONiC so far since no application uses tty mode redis-cli output
-        """
-        if resp is None:
-            print ""
-        elif isinstance(resp, list):
-            print "\n".join(resp)
+'''
+  Examples of how to use this script
```
dzhangalibaba, line 7: This helper msg is not showing anymore ?

judyjoseph: It shows the following output now, it is automatically generated by argparse library.
let me know the feedback, is it ok ?

admin@HOST:~$ sonic-db-cli -h
usage: sonic-db-cli [-h] [-n NAMESPACE] [database] cmd [cmd ...]

SONiC DB CLI

positional arguments:
  database              Database name
  cmd                   Command to execute in database

optional arguments:
  -h, --help            show this help message and exit
  -n NAMESPACE, --namespace NAMESPACE
                        Namespace string to use asic0/asic1.../asicn

judyjoseph:
> load_sonic_global_db_config will try to read /var/run/redis/sonic-db/database_global.json, but the file in at this location is generated via database docker, how this python lib works without changing database docker first ?

The API load_sonic_global_db_config() doesn't have a strict check for the existence of database_global.json. If the file is present it parses it. I didn't want to enforce .. as this file is only present/generated in the host database docker alone, needn't be present in all namespaces. Hence it is ok -- we can go ahead with this change without changing the database docker first.

src/swsssdk/scripts/sonic-db-cli
dzhangalibaba: prefer to keep the old usage and example message as well, probably can use epilog parameter for arguments to including the old string

dzhangalibaba:
My understanding is if the global config is not there, actually it means there is no namespace introduced and the old load func will load whatever we have today. If the global config exists, it load each namespace with it own config, then based on the namespace to choose the config. 

What I noticed is with current changes in this PR, those get API accept namespace parameter, if the global config doesn't exist and the API with namespace parameter is called , it will crash , correct ? I understand for non-namespace case it is fine. 

Also, if the global config exists, it load SonicDBConfig._sonic_db_config[''] twice ? one in load_sonic_db_config and other in load_sonic_global_db_config while doing namespace_validation ?


> The API load_sonic_global_db_config() doesn't have a strict check for the existence of database_global.json. If the file is present it parses it. I didn't want to enforce .. as this file is only present/generated in the host database docker alone, needn't be present in all namespaces. Hence it is ok -- we can go ahead with this change without changing the database docker first.

dzhangalibaba:
I added sonic-db-cli PING support and add one more get_instanceslist() API in SonicDBConfig.
I tried to keep your argparse format in this PR to make your resolve conflicts easily. You may want to do a rebase.

src/swsssdk/dbconnector.py
```diff
@@ -27,85 +78,129 @@ def load_sonic_db_config(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
...
             logger.exception(msg)
             raise RuntimeError(msg)
         SonicDBConfig._sonic_db_config_init = True
 
     @staticmethod
-    def db_name_validation(db_name):
+    def namespace_validation(namespace):
+        # Load global config if the namespace is an external one.
+        if namespace != '' and SonicDBConfig._sonic_db_global_config_init == False:
+            SonicDBConfig.load_sonic_global_db_config()
```
dzhangalibaba, line 93: one more thing: namespace_validation() is a library API, if user use namespace_validation() to check namaspace stuff before any load_sonic_db_config, it calls load_sonic_global_db_config() which actually load ns='' case, but didn't set _sonic_db_config_init = TURE, which will make following lines SonicDBConfig.load_sonic_db_config() still load ns='' case again, better to set _sonic_db_config_init = TURE in load_sonic_global_db_config() when ns is ' '.

src/swsssdk/dbconnector.py
```diff
@@ -9,10 +9,61 @@
...
+                    db_include_file = os.path.join(global_db_config_dir, entry['include'])
+    
+                    # Not finding the database_config.json file for the namespace
+                    if os.path.isfile(db_include_file) == False:
+                        msg = "'{}' file is not found !!".format(db_include_file)
+                        logger.warning(msg)
+                        continue
+
+                    with open(db_include_file, "r") as inc_file:
+                        SonicDBConfig._sonic_db_config[ns] = json.load(inc_file)
```
dzhangalibaba, line 63: if ns is ' ', better to set _sonic_db_config_init to avoid duplicated load.

src/swsssdk/scripts/sonic-db-cli
```diff
@@ -36,3 +59,37 @@ else:
...
+    parser.add_argument('db_or_op', type=str, help='Database name Or Unary operation(only PING supported)')
+    parser.add_argument('cmd', nargs='*', type=str, help='Command to execute in database')
+
+    args = parser.parse_args()
+
+    if args.db_or_op:
+        if args.cmd:
+            execute_cmd(args.db_or_op, args.cmd, args.namespace)
+        elif args.db_or_op == 'PING':
+            ping_all_instances()
```
dzhangalibaba, line 85: add namespace parameter to get correct instancelist ?

src/swsssdk/scripts/sonic-db-cli
```diff
@@ -1,30 +1,53 @@
...
+    rsp = False
+    msg = 'Could not connect to Redis at {}:{}: Connection refused'.format(inst_hostname, inst_port)
+    try:
+        rsp = r.ping()
+    except redis.exceptions.ConnectionError as e:
+        pass
+    return 'PONG' if rsp else msg
+
+def ping_all_instances():
+    db_insts = swsssdk.SonicDBConfig.get_instancelist()
```
dzhangalibaba, line 21: we need get the namesapce info for get_instancelist()

src/swsssdk/dbconnector.py
qiluo-msft: This is a great catch! In stead of make it smarter, how about making it more stupid and simple?

Since use are required to call `load_sonic_global_db_config()` at this use case, let's make this function throw runtime exception and educate user to call `load_sonic_global_db_config` first.

Also take Dong's suggestion to ensure `sonic_db_config_init == True` after calling `load_sonic_global_db_config()`


---
In reply to: [408568390](https://github.com/Azure/sonic-py-swsssdk/pull/63#discussion_r408568390) [](ancestors = 408568390)

src/swsssdk/scripts/sonic-db-cli
judyjoseph: shall update, I skipped earlier as these were new functions.

src/swsssdk/scripts/sonic-db-cli
judyjoseph: shall update, I skipped earlier as these were new functions.

src/swsssdk/dbconnector.py
judyjoseph: I am not 100% sure of this use case. The idea I had to make these two independent the load_global_config and local_db_config + to keep the existing behaviour of local_db_config intact.

I can set this flag sonic_db_config_init == True in load_global_config for ' ' case, **but then** the user cannot later if needed override the config present in global_config.json for the current namespace, with **a config which he wants later** .. by calling load_sonic_db_config(sonic_db_file_path=**my_config_file**)

src/swsssdk/dbconnector.py
judyjoseph: same as earlier comment , to keep these two independent, the load_global_config and local_db_config

src/swsssdk/scripts/sonic-db-cli
```diff
@@ -34,8 +34,11 @@ def ping_all_instances():
         print 'PONG'
         sys.exit(0)
 
-def execute_cmd(dbname, cmd):
-    dbconn = swsssdk.SonicV2Connector(use_unix_socket_path=False)
+def execute_cmd(dbname, cmd, namespace):
+    if namespace is None:
```
qiluo-msft, line 38: >if namespace is None: [](start = 4, length = 21)

If you change the command line arg `default=''`, then namespace will never be None, and you don't need to check and SonicV2Connector will handle it well.

src/swsssdk/dbconnector.py
dzhangalibaba: It should NOT be allowed to make user override existing config. It only should init once to keep consistence among all applications. Othewise some  application want this config , others want other config, that will be complex.

src/swsssdk/scripts/sonic-db-cli
judyjoseph: The check for namespace == None, is basically done to make sure we use the default TCP port when connecting to the DB in the local namespace and we maintain the current behavior. In case the user gives a namespace argument, we use the unix socket to connect, and when using the unix socket we would need to use "sudo". 

So the idea here is to make sure we use unix socket only when user specifies a namespace.

src/swsssdk/dbconnector.py
judyjoseph: I have updated with this flag sonic_db_config_init = True in the function load_sonic_global_db_config(). 

I haven't removed the implicit call to load_sonic_global_db_config() here if namespace is != '', as other wise the applications ( including scripts which uses it like sonic-db-cli, sonic_cfggen . db_migrator etc ) will have to call it explicitly  SonicDBConfig.load_sonic_global_db_config() ?

src/swsssdk/dbconnector.py
```diff
@@ -27,91 +84,129 @@ def load_sonic_db_config(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
...
             msg = "Could not open sonic database config file '{}'".format(sonic_db_file_path)
             logger.exception(msg)
             raise RuntimeError(msg)
         SonicDBConfig._sonic_db_config_init = True
 
     @staticmethod
-    def db_name_validation(db_name):
+    def namespace_validation(namespace):
+        # Load global config if the namespace is an external one.
+        if namespace != '' and SonicDBConfig._sonic_db_global_config_init == False:
```
qiluo-msft, line 98: >namespace [](start = 11, length = 9)

Do you need to check namespace is None, and raise exception at function beginning?

src/swsssdk/scripts/sonic-db-cli
```diff
@@ -34,8 +52,11 @@ def ping_all_instances():
...
```
qiluo-msft, line 56: >if namespace is None: [](start = 4, length = 21)

I see the intelligence to auto choosing a use_unix_socket_path based on use case. Thanks!

It will be useful to provide another command line argument `-s/--unix-socket` to let use override this choosing.

dzhangalibaba:
my earlier comments was addressed, looks good now from my side. You may want to address Qi's comment.

src/swsssdk/scripts/sonic-db-cli
judyjoseph: Updated with a new option.

src/swsssdk/dbconnector.py
judyjoseph: added the check

src/swsssdk/scripts/sonic-db-cli
```diff
@@ -1,11 +1,24 @@
...
+    r = redis.Redis(host=inst_hostname, unix_socket_path=inst_unix_socket_path)
+    rsp = False
+    msg = 'Could not connect to Redis at {}:{}: Connection refused'.format(inst_hostname, inst_unix_socket_path)
+    try:
+        rsp = r.ping()
+    except redis.exceptions.ConnectionError as e:
+        pass
+    return 'PONG' if rsp else msg
+
+def ping_tcp_single_instance(inst_info):
```
qiluo-msft, line 21: >ping_tcp_single_instance [](start = 4, length = 24)

Most code duplicated in ping_tcp_single_instance and ping_unix_path_single_instance. Not a blocking issue, and with you could improve them.

src/swsssdk/scripts/sonic-db-cli
```diff
@@ -1,11 +1,24 @@
 #!/usr/bin/python
+from __future__ import print_function
```
dzhangalibaba, line 2: just want to make sure if this PR rebased to the latest, I saw there is another commit yesterday with this changes, it should not be shown in your changes set.

judyjoseph: Ok, I too notice this, but I did a rebase and it was clean, not sure why it is showing it. Let me check this

judyjoseph: looks to me that there is no merge conflict, hence resolving it. Will make sure that merging is fine.

src/swsssdk/dbconnector.py
judyjoseph: I am taking Qi's suggestion to make the user explicitly call load_sonic_global_db_config(). 

In addition, I am optimizing the logic a bit by adding a new argument "namespace" to this API  load_sonic_global_db_config(). So if the user inputs an argument "namespace" this API will load the json file for that namespace alone .

This will improve performance of  utilitiles like sonic-cfggen/sonic-db-cli etc where we pass the namespace argument.

src/swsssdk/dbconnector.py
```diff
@@ -27,91 +93,136 @@ def load_sonic_db_config(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
...
+            raise RuntimeError(msg)
+        if SonicDBConfig._sonic_db_config_init == False:
+            SonicDBConfig.load_sonic_db_config()
+        if namespace not in SonicDBConfig._sonic_db_config:
+            msg = "{} is not a valid namespace name in configuration file".format(namespace)
+            logger.warning(msg)
+            raise RuntimeError(msg)
+
+    @staticmethod
+    def db_name_validation(db_name, namespace=''):
```
lguohan, line 124: minor comments, why not use namespace=None here?

src/swsssdk/dbconnector.py
```diff
@@ -27,91 +93,136 @@ def load_sonic_db_config(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
...
             msg = "Could not open sonic database config file '{}'".format(sonic_db_file_path)
             logger.exception(msg)
             raise RuntimeError(msg)
         SonicDBConfig._sonic_db_config_init = True
 
     @staticmethod
-    def db_name_validation(db_name):
+    def namespace_validation(namespace):
+        # Check the namespace is valid.
+        if namespace is None:
```
lguohan, line 107: why not treating None as global namespace?

src/swsssdk/config/database_global.json
```diff
@@ -0,0 +1,20 @@
+{
+    "INCLUDES" : [
```
lguohan, line 2: if this is for unit test, then it should be moved to tests/config folder

src/swsssdk/dbconnector.py
```diff
@@ -27,91 +93,136 @@ def load_sonic_db_config(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
                 logger.warning(msg)
                 sonic_db_file_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'config', 'database_config.json')
             with open(sonic_db_file_path, "r") as read_file:
-                SonicDBConfig._sonic_db_config = json.load(read_file)
+                # The database_config.json is loaded into the '' index, which referes to the local namespace.
```
lguohan, line 96: '' index seems a little bit confusing.

test/test_moduleLoad.py
```diff
@@ -13,3 +13,27 @@ def test__db_map_attributes(self):
         db = swsssdk.SonicV2Connector()
         self.assertTrue(all(hasattr(db, db_name) for db_name in db.get_db_list()))
         pass
```
lguohan, line 15: >pass [](start = 8, length = 4)

this pass is not necessary.

src/swsssdk/scripts/sonic-db-cli
```diff
@@ -51,34 +77,42 @@ def execute_cmd(dbname, cmd):
...
+
+Example 1: sonic-db-cli -n asic0 CONFIG_DB keys \*
+Example 2: sonic-db-cli -n asic2 APPL_DB HGETALL VLAN_TABLE:Vlan10
 Example 3: sonic-db-cli APPL_DB HGET VLAN_TABLE:Vlan10 mtu
-Example 4: sonic-db-cli APPL_DB EVAL "return {KEYS[1],KEYS[2],ARGV[1],ARGV[2]}" 2 k1 k2 v1 v2
+Example 4: sonic-db-cli -n asic3 APPL_DB EVAL "return {KEYS[1],KEYS[2],ARGV[1],ARGV[2]}" 2 k1 k2 v1 v2
 Example 5: sonic-db-cli PING
+Example 6: sonic-db-cli -s PING
 """)
+    parser.add_argument('-s', '--unixsocket', help="Override use of tcp_port and use unixsocket",action='store_true')
```
lguohan, line 101: >action [](start = 97, length = 6)

space here.

judyjoseph: Updated.

test/test_moduleLoad.py
judyjoseph: Updated.

src/swsssdk/dbconnector.py
judyjoseph: As discussed, changed the default value of namespace parameter as None. If the namespace is coming as None in any of the API's -- it is converted internally to the '' string referring to local namespace.

Here we check for namespace is None for a scenario where user calls this SonicDBConfig.namespace_validation() API explicitly and passes None. We error out in that case.

src/swsssdk/dbconnector.py
```diff
@@ -27,91 +93,154 @@ def load_sonic_db_config(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
...
+
+        if namespace not in SonicDBConfig._sonic_db_config:
+            msg = "{} is not a valid namespace name in configuration file".format(namespace)
+            logger.warning(msg)
+            raise RuntimeError(msg)
+
+    @staticmethod
+    def db_name_validation(db_name, namespace=None):
+        if namespace is None:
+            namespace = ''
```
lguohan, line 129: instead of repeating this, can you define a EMPTY_NAMESPACE and use it in the program.

judyjoseph: Defined a new function and use it to check and sent if namespace is None.

src/swsssdk/config/database_global.json
judyjoseph: Moved the file to tests/config

src/swsssdk/dbconnector.py
judyjoseph: Updated based on discussion.

src/swsssdk/dbconnector.py
judyjoseph: Updated based on discussion to continue using '' internally but externally the user will specify namespace=None, when he wants to connect to local instances.

rlhui:
@judyjoseph , there's conflict cherry-picking.  Please raise a separate PR for 201911 branch for this. Thanks.

judyjoseph:
> @judyjoseph , there's conflict cherry-picking. Please raise a separate PR for 201911 branch for this. Thanks.

This PR has dependencies with the changes done in the commits just before this (eg: https://github.com/Azure/sonic-py-swsssdk/commit/d1a15124719872c415cdaec680016e0035110ffe) , so cannot cherrypick this one alone... Is it possible to advance the sonic-py-swssdk submodule in 201911 ? I can raise a PR for advancing the submodule itself in 201911.

test/test_moduleLoad.py
```diff
@@ -12,4 +12,25 @@ def test__db_map_attributes(self):
...
+        filepath = os.path.join(os.path.dirname(os.path.abspath(__file__)), './config', 'database_global.json')
+        dbConfig.load_sonic_global_db_config(global_db_file_path=filepath)
+        ns_input = ['', 'asic0', 'asic1', 'asic2']
+        ns_list = list(dbConfig.get_ns_list())
+        ns_input.sort()
+        ns_list.sort()
+        self.assertEqual(ns_input, ns_list)
+
+    # This is the test to check if the global config file and get the correct DB in a namespace
+    def test__namespace_list(self):
```
qiluo-msft, line 29: Duplicated function name

src/swsssdk/dbconnector.py
```diff
@@ -9,10 +9,76 @@
...
+    """
+    @staticmethod
+    def load_sonic_global_db_config(global_db_file_path=SONIC_DB_GLOBAL_CONFIG_FILE, namespace=None):
+        """
+        Parse and load the global database config json file
+        """
+        if SonicDBConfig._sonic_db_global_config_init == True:
+            return
+
+        if os.path.isfile(global_db_file_path) ==  True:
```
qiluo-msft, line 34: >global_db_file_path [](start = 26, length = 19)

If global_db_file_path does not exist, why set _sonic_db_global_config_init to True?

Also can you return _sonic_db_global_config_init value to caller?

PR #68, opened 2020-04-28
[Multiasic]: Add calculation of interface index for backplane interfaces

SuvarnaMeenakshi:
Add calculation of interface index for backplane interfaces
in multi asic platform. Multiasic platform will have
backplane interface names of the format Ethernet-BP<x>.

Signed-off-by: SuvarnaMeenakshi <sumeenak@microsoft.com>

### DIFF
--- a/src/swsssdk/port_util.py
+++ b/src/swsssdk/port_util.py
@@ -6,6 +6,10 @@ import re
 
 
 SONIC_ETHERNET_RE_PATTERN = "^Ethernet(\d+)$"
+"""
+Ethernet-BP refers to BackPlane interfaces
+in multi-asic platform.
+"""
 SONIC_ETHERNET_BP_RE_PATTERN = "^Ethernet-BP(\d+)$"
 SONIC_PORTCHANNEL_RE_PATTERN = "^PortChannel(\d+)$"
 SONIC_MGMT_PORT_RE_PATTERN = "^eth(\d+)$"
--- a/src/swsssdk/port_util.py
+++ b/src/swsssdk/port_util.py
@@ -6,12 +6,14 @@ import re
 
 
 SONIC_ETHERNET_RE_PATTERN = "^Ethernet(\d+)$"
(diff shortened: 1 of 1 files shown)


src/swsssdk/port_util.py
```diff
@@ -6,19 +6,22 @@
 
 
 SONIC_ETHERNET_RE_PATTERN = "^Ethernet(\d+)$"
+SONIC_ETHERNET_BP_RE_PATTERN = "^Ethernet-BP(\d+)$"
```
qiluo-msft, line 9: >BP [](start = 42, length = 2)

Explain in comment what is BP?

src/swsssdk/port_util.py
```diff
@@ -6,19 +6,22 @@
...
 SONIC_PORTCHANNEL_RE_PATTERN = "^PortChannel(\d+)$"
 SONIC_MGMT_PORT_RE_PATTERN = "^eth(\d+)$"
 
 
 class BaseIdx:
     ethernet_base_idx = 1
+    ethernet_bp_base_idx = 9000
```
qiluo-msft, line 16: >9000 [](start = 27, length = 4)

How do you choose this value? Is 2000 working?

qiluo-msft: What is the range used by a realistic SKU?

---
In reply to: [416935612](https://github.com/Azure/sonic-py-swsssdk/pull/68#discussion_r416935612) [](ancestors = 416935612)

SuvarnaMeenakshi: Chose 9000 as the offset will be large enough and Index does not overlap. 
With 2000 as offset, it might overlap with PortChannel interface index.

prsunny: @qiluo-msft , i think 9000 should be a good safe offset value for back-plane instead of a lower value

src/swsssdk/port_util.py
SuvarnaMeenakshi: added as per comment

PR #71, opened 2020-05-21
Decode byte encoded response from redis before printing

rajendra-dendukuri:
Fix https://github.com/Azure/sonic-buildimage/issues/4632

In py3, the response from redis connector is encoded as byte array. They
need to be decoded before accessing them as strings.

Use following commands to test

sonic-db-cli CONFIG_DB "keys *"
sonic-db-cli CONFIG_DB "hget PORT|Ethernet0 admin_status"

Signed-off-by: Rajendra Dendukuri <rajendra.dendukuri@broadcom.com>

### DIFF
--- a/src/swsssdk/scripts/sonic-db-cli
+++ b/src/swsssdk/scripts/sonic-db-cli
@@ -7,6 +7,14 @@ import argparse
 from multiprocessing import Pool
 from functools import partial
 
+def print_response(resp):
+    if resp is None:
+        print()
+    elif isinstance(resp, list):
+        for x in resp: print(x.decode('utf-8') if isinstance(x, bytes) else x)
+    else:
+        print(resp.decode('utf-8') if isinstance(resp, bytes) else resp)
+
 def handle_single_instance(op, use_unix_socket, inst_info):
     inst_hostname = inst_info['hostname']
     if use_unix_socket:
@@ -76,12 +84,7 @@ def execute_cmd(dbname, cmd, namespace, use_unix_socket=False):
         based on our usage in SONiC, None and list type output from python API needs to be modified
         with these changes, it is enough for us to mimic redis-cli in SONiC so far since no application uses tty mode redis-cli output
         """
-        if resp is None:
-            print()
-        elif isinstance(resp, list):
-            print("\n".join(resp))
-        else:
-            print(resp)
+        print_response(resp)
         sys.exit(0)
 
 def main():
--- a/src/swsssdk/scripts/sonic-db-cli
+++ b/src/swsssdk/scripts/sonic-db-cli
@@ -7,14 +7,6 @@ import argparse
 from multiprocessing import Pool
 from functools import partial
 
-def print_response(resp):
-    if resp is None:
-        print()
-    elif isinstance(resp, list):
-        for x in resp: print(x.decode('utf-8') if isinstance(x, bytes) else x)
-    else:
-        print(resp.decode('utf-8') if isinstance(resp, bytes) else resp)
-
 def handle_single_instance(op, use_unix_socket, inst_info):
     inst_hostname = inst_info['hostname']
     if use_unix_socket:
@@ -65,11 +57,11 @@ def handle_all_instances(namespace, op, use_unix_socket=False):
 def execute_cmd(dbname, cmd, namespace, use_unix_socket=False):
     if namespace is None:
         if use_unix_socket:
-            dbconn = swsssdk.SonicV2Connector(use_unix_socket_path=True)
+            dbconn = swsssdk.SonicV2Connector(use_unix_socket_path=True, decode_responses=True)
         else:
-            dbconn = swsssdk.SonicV2Connector(use_unix_socket_path=False)
+            dbconn = swsssdk.SonicV2Connector(use_unix_socket_path=False, decode_responses=True)
     else:
-        dbconn = swsssdk.SonicV2Connector(use_unix_socket_path=True, namespace=namespace)
+        dbconn = swsssdk.SonicV2Connector(use_unix_socket_path=True, namespace=namespace, decode_responses=True)
     try:
         dbconn.connect(dbname)
     except RuntimeError:
@@ -84,7 +76,12 @@ def execute_cmd(dbname, cmd, namespace, use_unix_socket=False):
         based on our usage in SONiC, None and list type output from python API needs to be modified
         with these changes, it is enough for us to mimic redis-cli in SONiC so far since no application uses tty mode redis-cli output
         """
-        print_response(resp)
+        if resp is None:
+            print()
+        elif isinstance(resp, list):
+            print("\n".join(resp))
+        else:
+            print(resp)
         sys.exit(0)
 
 def main():


rajendra-dendukuri:
@dzhangalibaba  Can you please review this change.

src/swsssdk/scripts/sonic-db-cli
```diff
@@ -7,6 +7,14 @@ import argparse
 from multiprocessing import Pool
 from functools import partial
 
+def print_response(resp):
```
dzhangalibaba, line 10: do we need add similar changes to other python script ? Or only sonic-db-cli uses py3 today in SONiC ?

dzhangalibaba:
@qiluo-msft do you have other concern about this changes ?

dzhangalibaba:
issue seen without this changes.   https://github.com/Azure/sonic-buildimage/issues/4632

src/swsssdk/scripts/sonic-db-cli
rajendra-dendukuri: Those who are using the py3 version of py-swsssdk need to take care of decoding. sonic-db-cli happens to be one user. ZTP will be the other user of py3-swsssdk and in my yet to be raised PR's, I take of that as well.

src/swsssdk/scripts/sonic-db-cli
```diff
@@ -7,6 +7,14 @@ import argparse
...
+    if resp is None:
+        print()
+    elif isinstance(resp, list):
```
rajendra-dendukuri, line 13: @dzhangalibaba  Can you please comment on the below observation.

After more testing, I observed that when we give the below command, it is printing out the dictionary and not the list like in redis-cli

In py3
sonic-db-cli CONFIG_DB HGETALL "PORT|Ethernet0"
{b'index': b'0', b'alias': b'Ethernet1/0/1', b'admin_status': b'up', b'mtu': b'9100', b'speed': b'40000', b'lanes': b'1,2,3,4'}

In py2
root@sonic:/home/admin# sonic-db-cli CONFIG_DB HGETALL "PORT|Ethernet0"
{'index': '0', 'lanes': '1,2,3,4', 'mtu': '9100', 'alias': 'Ethernet1/0/1', 'admin_status': 'up', 'speed': '40000'}

There is a difference if we use hgetall
root@sonic:/home/admin# sonic-db-cli CONFIG_DB hgetall "PORT|Ethernet0"
index
0
lanes
1,2,3,4
mtu
9100
alias
Ethernet1/0/1
admin_status
up
speed
40000

qiluo-msft:
>         resp = client.execute_command(*cmd)

`redis.StrictRedis(decode_responses=True)` will handle the decoding in both python2 and python3. So suggest add extra parameter when contructing SonicV2Connector dbconn.

---
Refers to: src/swsssdk/scripts/sonic-db-cli:81 in a5ed19e. [](commit_id = a5ed19e887dbb8892105c90352c39bde369a1878, deletion_comment = False)

src/swsssdk/scripts/sonic-db-cli
dzhangalibaba: looks the redis-py has some different behavior when the input parameter are in different mformat. When quota "HGETALL PORT|Ethernet0", result is a list, otherwise is a dict . Interesting. The good thing is looks no application use HGETALL in current SONiC codes, it they want to use, need quote the whole command. Better to convert the dict to a list to avoid possible confuse. 

```
admin@ASW-7005:~$ sonic-db-cli CONFIG_DB HGETALL "PORT|Ethernet0"
{'alias': 'Ethernet1/1', 'admin_status': 'up', 'lanes': '17', 'index': '0', 'mtu': '9100'}
admin@ASW-7005:~$ sonic-db-cli  CONFIG_DB "HGETALL PORT|Ethernet0"
alias
Ethernet1/1
admin_status
up
mtu
9100
lanes
17
index
0

```

could you add something like this, then I don't need to raise another PR.
```
def print_response(resp):

    if isinstance(resp, dict):
        resp = list(sum(zip(resp.keys(),resp.values()),()))

    if resp is None:
        print()
    elif isinstance(resp, list):
        for x in resp: print(x.decode('utf-8') if isinstance(x, bytes) else x)
    else:
        print(resp.decode('utf-8') if isinstance(resp, bytes) else resp)

```

rajendra-dendukuri:
> > ```
> >     resp = client.execute_command(*cmd)
> > ```
> 
> `redis.StrictRedis(decode_responses=True)` will handle the decoding in both python2 and python3. So suggest add extra parameter when contructing SonicV2Connector dbconn.
> 
> Refers to: src/swsssdk/scripts/sonic-db-cli:81 in [a5ed19e](https://github.com/Azure/sonic-py-swsssdk/commit/a5ed19e887dbb8892105c90352c39bde369a1878). [](commit_id = [a5ed19e](https://github.com/Azure/sonic-py-swsssdk/commit/a5ed19e887dbb8892105c90352c39bde369a1878), deletion_comment = False)

Thanks @qiluo-msft  I finished my changes and pushed them. Below is the output with my changes.

root@sonic:/home/admin# ./test.sh
++ sonic-db-cli CONFIG_DB 'KEYS PORT|Ethernet1*'
PORT|Ethernet116
PORT|Ethernet16
PORT|Ethernet104
PORT|Ethernet112
PORT|Ethernet124
PORT|Ethernet108
PORT|Ethernet100
PORT|Ethernet120
PORT|Ethernet12
++ sonic-db-cli CONFIG_DB 'HGETALL PORT|Ethernet0'
index
0
lanes
1,2,3,4
mtu
9100
alias
Ethernet1/0/1
admin_status
up
speed
40000
++ sonic-db-cli CONFIG_DB 'hgetall PORT|Ethernet0'
index
0
lanes
1,2,3,4
mtu
9100
alias
Ethernet1/0/1
admin_status
up
speed
40000
++ sonic-db-cli CONFIG_DB 'HGET PORT|Ethernet0 admin_status'
up
root@sonic:/home/admin#

src/swsssdk/scripts/sonic-db-cli
rajendra-dendukuri: With latest changes to use decode_responses=True in the dbconn, everything shows up right. Both hgetall and HGETALL. 

root@sonic:/home/admin# sonic-db-cli CONFIG_DB "hgetall PORT|Ethernet0"
index
0
lanes
1,2,3,4
mtu
9100
alias
Ethernet1/0/1
admin_status
up
speed
40000
root@sonic:/home/admin# sonic-db-cli CONFIG_DB "HGETALL PORT|Ethernet0"
index
0
lanes
1,2,3,4
mtu
9100
alias
Ethernet1/0/1
admin_status
up
speed
40000
root@sonic:/home/admin#

qiluo-msft: @dzhangalibaba Reading the code of `redis/client.py` I feel we need to call `c.execute_command('hgetall', 'PORT|Ethernet0')` instead of `c.execute_command('hgetall PORT|Ethernet0')`, because `execute_command` will use `arg[0]` to post-parse response. We need to consider this and limit in `sonic-db-cli`. Could you raise another PR to fix them?

dzhangalibaba:
@qiluo-msft @rajendra-dendukuri  do we need to add default decode_responses=True in library SonicV2Connector init() instead of the application setting it?  In future, if other application change to py3 codes, this may need to be add in the every application again.

qiluo-msft:
In future this is a good idea. This PR is just a quick bug fix, so it is enough now and well tested.

Issue #72, opened 2020-05-22
Exception seen when trying to connect and database service is not responding


rajendra-dendukuri:
1. Stop all services including database and verify that none of the containers are running.

root@sonic:/home/admin# docker ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
root@sonic:/home/admin#

2 Now try to initiate a connection with retry option. A py exception is observed.

root@sonic:/home/admin# cat test.py
#!/usr/bin/env python2.7
from swsssdk import SonicV2Connector

state_db = SonicV2Connector(host='17.0.0.1')
state_db.connect(state_db.STATE_DB,True)
root@sonic:/home/admin#

root@sonic:/home/admin# ./test.py
Traceback (most recent call last):
  File "./test.py", line 5, in <module>
    state_db.connect(state_db.STATE_DB,True)
  File "/usr/local/lib/python2.7/dist-packages/swsssdk/dbconnector.py", line 255, in connect
    super(SonicV2Connector, self).connect(db_id, retry_on)
  File "/usr/local/lib/python2.7/dist-packages/swsssdk/interface.py", line 169, in connect
    self._persistent_connect(db_id)
  **File "/usr/local/lib/python2.7/dist-packages/swsssdk/interface.py", line 198, in _persistent_connect
    self.close(db_id)**
  File "/usr/local/lib/python2.7/dist-packages/swsssdk/dbconnector.py", line 258, in close
    db_id = self.get_dbid(db_name)
  File "/usr/local/lib/python2.7/dist-packages/swsssdk/dbconnector.py", line 277, in get_dbid
    return SonicDBConfig.get_dbid(db_name, self.namespace)
  File "/usr/local/lib/python2.7/dist-packages/swsssdk/dbconnector.py", line 215, in get_dbid
    SonicDBConfig.db_name_validation(db_name, namespace)
  File "/usr/local/lib/python2.7/dist-packages/swsssdk/dbconnector.py", line 143, in db_name_validation
    raise RuntimeError(msg)
RuntimeError: 6 is not a valid database name in configuration file
root@sonic:/home/admin#

rajendra-dendukuri:
@dzhangalibaba  Can you please take a look at this.

diff --git a/src/swsssdk/interface.py b/src/swsssdk/interface.py
index a3a13cb..fa0d530 100644
--- a/src/swsssdk/interface.py
+++ b/src/swsssdk/interface.py
[...]

PR #74, opened 2020-05-22
[MultiDB] SonicV2Connector class use DBInterface class as member instead of inheritance

dzhangalibaba:
* issue : https://github.com/Azure/sonic-py-swsssdk/issues/72

* the root cause is : SonicV2Connector connect(dbname) calls base class method connect(dbid), while when connection failed and retry, base class try to call self. connect() which is onicV2Connector connect(dbname) since name are the same. It should call its own connect(dbid).

* for multiDB, we use methods from SonicV2Connector instead of DBInterface based on earlier changes

* we should make sure DBInterface class method uses its own methods inside, hence making DBInterface class as a member of SonicV2Connector instead of inheritance

### DIFF
--- a/src/swsssdk/dbconnector.py
+++ b/src/swsssdk/dbconnector.py
@@ -223,9 +223,9 @@ class SonicDBConfig(object):
         SonicDBConfig.db_name_validation(db_name, namespace)
         return SonicDBConfig._sonic_db_config[namespace]["DATABASES"][db_name]["separator"]
 
-class SonicV2Connector(DBInterface):
+class SonicV2Connector(object):
     def __init__(self, use_unix_socket_path=False, namespace=None, **kwargs):
-        super(SonicV2Connector, self).__init__(**kwargs)
+        self.dbintf = DBInterface(**kwargs)
         self.use_unix_socket_path = use_unix_socket_path
 
         """If the user don't give the namespace as input, it refers to the local namespace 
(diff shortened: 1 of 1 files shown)


dzhangalibaba:
@rajendra-dendukuri @qiluo-msft  Please help review

dzhangalibaba:
Test

When stop all dockers , retry connect until database is restarted and set/get works fine

```
admin@ASW-7005:~$ cat /etc/sonic/testdb.py 
#!/usr/bin/env python2.7
from swsssdk import SonicV2Connector

db = SonicV2Connector()
db.connect(db.CONFIG_DB,True)

db.set("CONFIG_DB", "hashhash", "kk", "vv")
print(db.get_all("CONFIG_DB", "hashhash"))
admin@ASW-7005:~$ /etc/sonic/testdb.py 
{'kk': 'vv'}
admin@ASW-7005:~$ 


admin@ASW-7005:~$ docker stop $(docker ps -aq)
edfd0e02be03
b1a76602159b
f16090b43b43
9b67c63b2c8f
bb2f217daea0
4026b5e72382
c090c9da180c
7ddce00e4387
23df59d0f92b
a7fc3d31a590
4d142775e9b3
5ee83628e713
48a5d88bdf5e
725d3b951e66
admin@ASW-7005:~$ /etc/sonic/testdb.py 
{'kk': 'vv'}
admin@ASW-7005:~$ 

```

src/swsssdk/interface.py
```diff
@@ -359,6 +359,6 @@ def _connection_error_handler(self, db_id):
         In the event Redis is unavailable, close existing connections, and try again.
         """
         logger.warning('Could not connect to Redis--waiting before trying again.')
-        self.close(db_id)
+        self._close(db_id)
         time.sleep(self.CONNECT_RETRY_WAIT_TIME)
-        self.connect(db_id, True)
+        self._connect(db_id, True)
```
qiluo-msft, line 364: >_connect [](start = 13, length = 8)

We need to fix SonicV2Connector. There is nothing wrong in DBInterface

qiluo-msft:
> class SonicV2Connector(DBInterface):

Could it use composition instead of inheritance?

---
Refers to: src/swsssdk/dbconnector.py:226 in ac457fb. [](commit_id = ac457fb90c0eb145a408fad3ce0bc6a2ed14518f, deletion_comment = False)

src/swsssdk/interface.py
dzhangalibaba: use composition instead and only change SonicV2Connector wrapper.

dzhangalibaba:
> > class SonicV2Connector(DBInterface):
> 
> Could it use composition instead of inheritance?
> 
> Refers to: src/swsssdk/dbconnector.py:226 in [ac457fb](https://github.com/Azure/sonic-py-swsssdk/commit/ac457fb90c0eb145a408fad3ce0bc6a2ed14518f). [](commit_id = [ac457fb](https://github.com/Azure/sonic-py-swsssdk/commit/ac457fb90c0eb145a408fad3ce0bc6a2ed14518f), deletion_comment = False)

yes we can use it. Updated

rajendra-dendukuri:
Verified the patch and it looks good.

SuvarnaMeenakshi:
@qiluo-msft , @dzhangalibaba is there a plan to merge this to 201911 branch?

qiluo-msft:
Yes. Added the label.

PR #77, opened 2020-05-31
Updates to the namespace API's for test_moduleLoad.py

judyjoseph:
Why
-----
Updates to the python module testcases for the namespaces. The changes include

* Add a database_config.json file for the test scripts in test/config directory. 
* Update the name of a test API which was missed out earlier 
* Changes in database_global.json file to test various input formats of database_config.json files like the absolute path, relative path etc.

### DIFF
--- a/test/config/database_global.json
+++ b/test/config/database_global.json
@@ -5,7 +5,7 @@
         },
         {
             "namespace" : "asic0",
-            "include" : "/sonic/src/sonic-py-swsssdk/test/config/database_config.json"
+            "include" : "../config/database_config.json"
         },
         {
             "namespace" : "asic1",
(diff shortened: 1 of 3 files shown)


test/config/database_global.json
```diff
@@ -5,11 +5,11 @@
         },
         {
             "namespace" : "asic0",
-            "include" : "database_config.json"
+            "include" : "/sonic/src/sonic-py-swsssdk/test/config/database_config.json"
```
qiluo-msft, line 8: >/sonic/src/sonic-py-swsssdk/test/config/ [](start = 25, length = 40)

The test may be run outside slave docker, so the absolute path may be invalid. Could you use relative path?

judyjoseph: Updated.

PR #78, opened 2020-06-01
[port_util] add get_rif_port_map, get_vlan_interface_oid_map

mykolaf:
Add 2 new utility functions, to be used by snmpagent.

Related to [#133](https://github.com/Azure/sonic-snmpagent/pull/133)

### DIFF
--- a/src/swsssdk/port_util.py
+++ b/src/swsssdk/port_util.py
@@ -11,12 +11,14 @@ Ethernet-BP refers to BackPlane interfaces
 in multi-asic platform.
 """
 SONIC_ETHERNET_BP_RE_PATTERN = "^Ethernet-BP(\d+)$"
+SONIC_VLAN_RE_PATTERN = "^Vlan(\d+)$"
 SONIC_PORTCHANNEL_RE_PATTERN = "^PortChannel(\d+)$"
 SONIC_MGMT_PORT_RE_PATTERN = "^eth(\d+)$"
 
 
 class BaseIdx:
     ethernet_base_idx = 1
+    vlan_interface_base_idx = 2000
     ethernet_bp_base_idx = 9000
     portchannel_base_idx = 1000
     mgmt_port_base_idx = 10000
@@ -25,6 +27,7 @@ def get_index(if_name):
     """
     OIDs are 1-based, interfaces are 0-based, return the 1-based index
     Ethernet N = N + 1
+    Vlan N = N + 2000
     Ethernet_BP N = N + 9000
     PortChannel N = N + 1000
     eth N = N + 10000
@@ -36,6 +39,7 @@ def get_index_from_str(if_name):
     """
     OIDs are 1-based, interfaces are 0-based, return the 1-based index
     Ethernet N = N + 1
+    Vlan N = N + 2000
     Ethernet_BP N = N + 9000
     PortChannel N = N + 1000
     eth N = N + 10000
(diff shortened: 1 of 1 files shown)


src/swsssdk/port_util.py
```diff
@@ -11,12 +11,14 @@
...
 """
 SONIC_ETHERNET_BP_RE_PATTERN = "^Ethernet-BP(\d+)$"
+SONIC_VLAN_RE_PATTERN = "^Vlan(\d+)$"
 SONIC_PORTCHANNEL_RE_PATTERN = "^PortChannel(\d+)$"
 SONIC_MGMT_PORT_RE_PATTERN = "^eth(\d+)$"
 
 
 class BaseIdx:
     ethernet_base_idx = 1
+    vlan_interface_base_idx = 2000
```
qiluo-msft, line 21: @prsunny @judyjoseph do you think it leaves enough indexes to portchannels in multiple-ASIC case?

src/swsssdk/port_util.py
```diff
@@ -103,3 +108,40 @@ def get_vlan_id_from_bvid(db, bvid):
...
+
+    return rif_port_oid_map
+
+def get_vlan_interface_oid_map(db):
+    """
+        Get Vlan Interface names and sai oids
+    """
+    db.connect('COUNTERS_DB')
+    rif_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_RIF_NAME_MAP', blocking=True)
+    rif_type_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_RIF_TYPE_MAP', blocking=True)
```
qiluo-msft, line 136: also check None?

src/swsssdk/port_util.py
```diff
@@ -103,3 +108,40 @@ def get_vlan_id_from_bvid(db, bvid):
...
+
+    if not rif_name_map:
+        return {}
+
+    oid_pfx = len("oid:0x")
+    vlan_if_name_map = {sai_oid[oid_pfx:]: if_name for if_name, sai_oid in rif_name_map.items()
```
qiluo-msft, line 142: This statement is too long to understand. Could you refactor to a loop with if-block? And add some comment to explain?

src/swsssdk/port_util.py
judyjoseph: I feel 1000 port channel interfaces should be good enough even for Chassis. But I am looking to see why the vlan_interface_base_idx is defined  .. is it for this API get_vlan_interface_oid_map() ?

prsunny: I think this just simply leaves a gap from 6000-9000 after `2000+4094`, why not start from 4000?

src/swsssdk/port_util.py
mykolaf: Sure

src/swsssdk/port_util.py
mykolaf: 2K BaseIdx is what our Arch and Guohan agreed on. If there is any concern about the size of pool, should discuss with them.

mykolaf: @judyjoseph 
>  is it for this API get_vlan_interface_oid_map() ?
Theoretically index can be used anywhere where a numeric identification for interface is needed. 
In my case I am using it for interface ID in the SNMP Interface MIB (RFC1213).

judyjoseph: lgtm .. I see that the index range discussion already happened.

PR #80, opened 2020-06-09
Upgrade redis library to 3.5.3

qiluo-msft:
Upgrade outdated redis python library to latest version, which is still supporting python2.


yxieca:
Can you add some description regarding the motivation of upgradeing?

PR #82, opened 2020-06-25
added abstraction for bulk operations through redis pipeline

anilkpan:
For new commands to add/del range of vlans and range of vlan members, added new interfaces for bulk redis pipeline update

### DIFF
--- a/src/swsssdk/configdb.py
+++ b/src/swsssdk/configdb.py
@@ -193,6 +193,42 @@ class ConfigDBConnector(SonicV2Connector):
                     k = k + '@'
                 client.hdel(_hash, self.serialize_key(k))
 
+    def set_bulk(self, payload):
+        """Write bulk entries to config db.
+        """
+        client = self.redis_clients[self.db_name]
+        pipe = client.pipeline()
+        for (k,v) in payload:
+            pipe.hmset(k, v)
+        pipe.execute()
+
+    def del_bulk(self, payload):
+        """Delete bulk entries from config db.
+        """
+        client = self.redis_clients[self.db_name]
+        pipe = client.pipeline()
+        for (k) in payload:
+            pipe.delete(k)
+        pipe.execute()
+
+    def hdel_bulk(self, payload):
+        """Delete bulk entries from config db.
+        """
+        client = self.redis_clients[self.db_name]
+        pipe = client.pipeline()
+        for (k,v) in payload:
+            pipe.hdel(k,v)
+        pipe.execute()
+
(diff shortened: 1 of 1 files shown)


src/swsssdk/configdb.py
```diff
@@ -193,6 +193,42 @@ def set_entry(self, table, key, data):
                     k = k + '@'
                 client.hdel(_hash, self.serialize_key(k))
 
+    def set_bulk(self, payload):
+        """Write bulk entries to config db.
+        """
+        client = self.redis_clients[self.db_name]
```
rajendra-dendukuri, line 199: You need to use the below API which is Multi-DB compliant.

self.get_redis_client(self.db_name). Modify other occurrences as well.

rajendra-dendukuri:
@qiluo-msft can you please review this change. These bulk API's improve performance when performing large number of table operations.

src/swsssdk/configdb.py
```diff
@@ -193,6 +193,42 @@ def set_entry(self, table, key, data):
                     k = k + '@'
                 client.hdel(_hash, self.serialize_key(k))
 
+    def set_bulk(self, payload):
```
qiluo-msft, line 196: >set_bulk [](start = 8, length = 8)

What are the use cases for the new functions? Application can achieve pipeline by calling get_redis_client().pipeline. And that solution will be more flexible since you can mix functions into the same pipeline.

rajendra-dendukuri: Qi, Please refer to the below comment from. It was suggested not to use the pipeline in the application.

https://github.com/Azure/sonic-utilities/pull/891#discussion_r420986191

The use case is also mentioned in the same PR. It is to create large number of VLANs. This also helps when querying and displaying the contents of a filled-up mac address table.

rajendra-dendukuri:
> Ask for use cases.
Below is a usecase.
https://github.com/Azure/sonic-utilities/pull/891

src/swsssdk/configdb.py
qiluo-msft: Thanks @rajendra-dendukuri sharing the background! As we planed to move away from sonic-py-swsssdk and converge everything to sonic-swss-common, I believe new feature here will be short-lived. We have implemented some pipeline support in sonic-swss-common *Tables classes. Could you check if that fulfill your requirement?

Sorry for the inconvenience to extend code here.

---
In reply to: [445300357](https://github.com/Azure/sonic-py-swsssdk/pull/82#discussion_r445300357) [](ancestors = 445300357)

PR #83, opened 2020-07-22
[configdb] Add Ability to Query/Update Redis Using Pipelines

tahmed-dev:
Redis recommend using pipeline in order to obtain optimal speen when
handling large volume of data. Te pipeline API also given an ability
devide work in batches.

singed-off-by: Tamer Ahmed <tamer.ahmed@microsoft.com>

Verified the new class using custom cfggen and saw the following performance gains.

Current Connector
```
admin@str-s6000-acs-14:~$ time sudo sonic-cfggen -j /etc/sonic/config_db.json --write-to-db

real	0m2.286s
user	0m1.608s
sys	0m0.364s
admin@str-s6000-acs-14:~$ time sonic-cfggen -d --print-data > dump_old_debug.json 

real	0m2.219s
user	0m1.816s
sys	0m0.360s
```

Pipe Connector
```
admin@str-s6000-acs-14:~$ time sudo sonic-cfggen -j /etc/sonic/config_db.json --write-to-db

real	0m1.648s
user	0m1.316s
sys	0m0.304s
admin@str-s6000-acs-14:~$ time sonic-cfggen -d --print-data > dump_new_debug.json 

real	0m2.006s
user	0m1.732s
sys	0m0.260s

```

### DIFF
--- a/src/swsssdk/configdb.py
+++ b/src/swsssdk/configdb.py
@@ -415,7 +415,7 @@ class ConfigDBPipeConnector(ConfigDBConnector):
         pipe.execute()
         client.bgsave()
 
-    def _get_config(self, client, pipe, data, cursor):
+    def __get_config(self, client, pipe, data, cursor):
         """Read config data in batches of size REDIS_SCAN_BATCH_SIZE using Redis pipelines
         Args:
             client: Redis client
@@ -454,9 +454,9 @@ class ConfigDBPipeConnector(ConfigDBConnector):
         pipe = client.pipeline()
         data = {}
 
-        cur = self._get_config(client, pipe, data, 0)
+        cur = self.__get_config(client, pipe, data, 0)
         while cur != 0:
-            cur = self._get_config(client, pipe, data, cur)
+            cur = self.__get_config(client, pipe, data, cur)
 
         return data
 
--- a/src/swsssdk/configdb.py
+++ b/src/swsssdk/configdb.py
@@ -120,7 +120,7 @@ class ConfigDBConnector(SonicV2Connector):
         for raw_key in raw_data:
             key = raw_key
             if PY3K:
-                key = raw_key.decode('utf-8')
+                key = raw_key.decode()
 
             # "NULL:NULL" is used as a placeholder for objects with no attributes
             if key == "NULL":
@@ -136,7 +136,7 @@ class ConfigDBConnector(SonicV2Connector):
                 typed_data[key[:-1]] = value
             else:
                 if PY3K:
-                    typed_data[key] = raw_data[raw_key].decode('utf-8')
+                    typed_data[key] = raw_data[raw_key].decode()
                 else:
                     typed_data[key] = raw_data[raw_key]
         return typed_data
@@ -240,7 +240,7 @@ class ConfigDBConnector(SonicV2Connector):
         for key in keys:
             try:
                 if PY3K:
-                    key = key.decode('utf-8')
+                    key = key.decode()
                 if split:
                     (_, row) = key.split(self.TABLE_NAME_SEPARATOR, 1)
                     data.append(self.deserialize_key(row))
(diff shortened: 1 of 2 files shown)


src/swsssdk/configdb.py
```diff
@@ -328,10 +328,135 @@ def get_config(self):
...
-                entry = self.__raw_to_typed(client.hgetall(key))
+                entry = self.raw_to_typed(client.hgetall(key))
                 if entry != None:
                     data.setdefault(table_name, {})[self.deserialize_key(row)] = entry
             except ValueError:
                 pass    #Ignore non table-formated redis entries
         return data
 
+
+class ConfigDBPipeConnector(ConfigDBConnector):
```
qiluo-msft, line 339: >ConfigDBPipeConnector [](start = 6, length = 21)

From the design point of view, ConfigDBPipeConnector is not a subclass of ConfigDBConnector (ConfigDBNonPipeConnector).

I suggest reuse original class.
```
def __init__(self, use_pipeline=False, **kwargs):
```

src/swsssdk/configdb.py
```diff
@@ -328,10 +328,135 @@ def get_config(self):
...
+            client: Redis client
+            pipe: Redis DB pipe
+            data: config dictionary
+            cursor: position to start scanning from
+
+        Returns:
+            cur: poition of next item to scan
+        """
+        cur, keys = client.scan(cursor=cursor, match='*', count=self.REDIS_SCAN_BATCH_SIZE)
+        keys = [key.decode('utf-8') for key in keys if key != self.INIT_INDICATOR]
```
qiluo-msft, line 430: >key.decode('utf-8') [](start = 16, length = 19)

You can just `key.decode()`

src/swsssdk/configdb.py
tahmed-dev: it is extending ConfigDBConnector functionality by providing pipeline based communication for mod_config, get_config. the use_pipeline switch resemble the use_unix_socket_path option. However, the use_pipeline is propagating to the connector class. This will result in if statement or maintaining internal vptr table which is done by subclassing in this case.

src/swsssdk/configdb.py
tahmed-dev: Ack

src/swsssdk/configdb.py
qiluo-msft: > However, the use_pipeline is propagating to the connector class

This is not true. you can check use_pipeline first and remove it from kwargs.

---
In reply to: [459153837](https://github.com/Azure/sonic-py-swsssdk/pull/83#discussion_r459153837) [](ancestors = 459153837)

tahmed-dev: Correct. What I meant is that use_pipeline switch is not the same as use_unix_socket _domain. It is use in this class will require if statement or creating vptr table which is given for free when extend the class.

src/swsssdk/configdb.py
```diff
@@ -325,13 +325,138 @@ def get_config(self):
...
+            client: Redis client
+            pipe: Redis DB pipe
+            table: Table name.
+        """
+        pattern = '{}{}*'.format(table.upper(), self.TABLE_NAME_SEPARATOR)
+        cur = self.__delete_entries(client, pipe, pattern, 0)
+        while cur != 0:
+            cur = self.__delete_entries(client, pipe, pattern, cur)
+
+    def __mod_entry(self, pipe, table, key, data):
```
qiluo-msft, line 377: >__mod_entry [](start = 8, length = 11)

why some functions start with double underscore, but some with one? Suggest use only one.

tahmed-dev: The module is using __ for private and so I maintained it. The function _get_config was an outlier and got fixed by recent push.

PR #87, opened 2020-09-29
[swsssdk] update DBInterface redis_client index from db_id to db_name

dzhangalibaba:
* DBInterface stores redis_client with index db_name

### DIFF
--- a/src/swsssdk/interface.py
+++ b/src/swsssdk/interface.py
@@ -18,7 +18,7 @@ def blockable(f):
 
         class SonicV2Connector:
             @blockable
-            def keys(self, db_id):
+            def keys(self, db_name):
                 # ...
 
         # call with:
@@ -181,11 +181,12 @@ class DBInterface(object):
         if db_name is None:
             raise ValueError("No database Name configured for '{}'".format(db_name))
 
-        client = redis.StrictRedis(db=db_id, **self.redis_kwargs)
+        if db_name not in self.redis_clients.keys():
+            client = redis.StrictRedis(db=db_id, **self.redis_kwargs)
 
-        # Enable the notification mechanism for keyspace events in Redis
-        client.config_set('notify-keyspace-events', self.KEYSPACE_EVENTS)
-        self.redis_clients[db_name] = client
+            # Enable the notification mechanism for keyspace events in Redis
+            client.config_set('notify-keyspace-events', self.KEYSPACE_EVENTS)
+            self.redis_clients[db_name] = client
 
     def _persistent_connect(self, db_id, db_name):
         """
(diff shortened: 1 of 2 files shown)


dzhangalibaba:
@qiluo-msft

src/swsssdk/interface.py
```diff
@@ -159,206 +159,210 @@ def __init__(self, **kwargs):
...
         """
         Connect to database id.
         """
         if db_id is None:
             raise ValueError("No database ID configured for '{}'".format(db_id))
 
+        if db_name is None:
+            raise ValueError("No database Name configured for '{}'".format(db_name))
+
         client = redis.StrictRedis(db=db_id, **self.redis_kwargs)
```
qiluo-msft, line 184: >redis.StrictRedis [](start = 17, length = 17)

If `self.redis_clients` already has this StrictRedis, we don't need to create a new one.

qiluo-msft:
>             def keys(self, db_id):

db_name

---
Refers to: src/swsssdk/interface.py:21 in 00f0543. [](commit_id = 00f05432785dae4e8bdf6a0b43bfc2787699cb47, deletion_comment = False)

qiluo-msft (changes requested):
As comments

dzhangalibaba:
> > ```
> >         def keys(self, db_id):
> > ```
> 
> db_name
> 
> Refers to: src/swsssdk/interface.py:21 in [00f0543](https://github.com/Azure/sonic-py-swsssdk/commit/00f05432785dae4e8bdf6a0b43bfc2787699cb47). [](commit_id = [00f0543](https://github.com/Azure/sonic-py-swsssdk/commit/00f05432785dae4e8bdf6a0b43bfc2787699cb47), deletion_comment = False)

UPDATED

src/swsssdk/interface.py
dzhangalibaba: it is a day 1 behavior, we might consider update all other previous release as well. Anyway we can fix it in dev-master for now. code updated.

qiluo-msft (approved):
Thanks!

src/swsssdk/interface.py
```diff
@@ -159,206 +159,211 @@ def __init__(self, **kwargs):
...
+    def _connection_error_handler(self, db_name):
         """
         In the event Redis is unavailable, close existing connections, and try again.
         """
         logger.warning('Could not connect to Redis--waiting before trying again.')
-        self.close(db_id)
+        self.close(db_name)
         time.sleep(self.CONNECT_RETRY_WAIT_TIME)
-        self.connect(db_id, True)
+        self.connect(db_name, True)
```
qiluo-msft, line 369: >self.connect [](start = 8, length = 12)

Number of parameters does not match.

imzyxwvu:
Hello! I saw that you wrapped the statements that turn on `notify-keyspace-events` in `_onetime_connect` from `interface.py` with a `db_name not in self.redis_clients.keys()` codition. Is there any reason behind this change?

PR #88, opened 2020-10-01
[MultiDB] fix wrong parameter number in interface.py

dzhangalibaba:
* correct wrong connect() parameter
* store and get db_id from local
* del old/unused redis and pubsub obj in array

### DIFF
--- a/src/swsssdk/interface.py
+++ b/src/swsssdk/interface.py
@@ -366,4 +366,4 @@ class DBInterface(object):
         logger.warning('Could not connect to Redis--waiting before trying again.')
         self.close(db_name)
         time.sleep(self.CONNECT_RETRY_WAIT_TIME)
-        self.connect(db_name, True)
+        self.connect(self.redis_clients[db_name].connection_pool.connection_kwargs['db'], db_name, True)
--- a/src/swsssdk/interface.py
+++ b/src/swsssdk/interface.py
@@ -155,6 +155,9 @@ class DBInterface(object):
         # Create a separate client for each database
         self.redis_clients = DBRegistry()
 
+        # record db_name to db_id mapping on local
+        self.redis_db_map = {}
+
         # Create a channel for receiving needed keyspace event
         # notifications for each client
         self.keyspace_notification_channels = DBRegistry()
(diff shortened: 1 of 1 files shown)


dzhangalibaba:
@qiluo-msft

src/swsssdk/interface.py
```diff
@@ -366,4 +370,4 @@ def _connection_error_handler(self, db_name):
         logger.warning('Could not connect to Redis--waiting before trying again.')
         self.close(db_name)
         time.sleep(self.CONNECT_RETRY_WAIT_TIME)
-        self.connect(db_name, True)
+        self.connect(self.redis_db_map[db_name], db_name, True)
```
qiluo-msft, line 373: Without a new dict, you can get `db_id` from `StrictRedis.connection_pool.connection_kwargs['db']`

qiluo-msft (commented):
As comment

src/swsssdk/interface.py
dzhangalibaba: updated

src/swsssdk/interface.py
```diff
@@ -366,4 +366,4 @@ def _connection_error_handler(self, db_name):
         logger.warning('Could not connect to Redis--waiting before trying again.')
         self.close(db_name)
         time.sleep(self.CONNECT_RETRY_WAIT_TIME)
-        self.connect(db_name, True)
+        self.connect(self.redis_clients[db_name].connection_pool.connection_kwargs['db'], db_name, True)
```
qiluo-msft, line 369: >self.redis_clients[db_name] [](start = 21, length = 27)

It is already closed by Line 367. So it is not in the array.

dzhangalibaba: close only do a disconnect, the redis obj is still there, I think that's a minor bug, we suppose to remove old obj after close. fix it in update and still using local dict to store db_name & db_id mapping

src/swsssdk/interface.py
qiluo-msft: Offline discussed. My proposal is based on implementation details. So let's keep your code.

qiluo-msft (approved):

PR #89, opened 2020-10-22
Make functions also working for SonicV2Connector with decode_responses=True

qiluo-msft:
More context:

We plan to migrate existing SonicV2Connector from `decode_responses=False` (default) to `decode_responses=True`
- [ ] this PR
- [ ] sonic-snmpagent uses only `decode_responses=True`, and remove all the `encode()`/`decode()`
- [ ] other applications with `encode()`/`decode()`: uses only `decode_responses=True`, and remove all the `encode()`/`decode()`
   - [ ] nbrshow
   - [ ] fdbshow
- [ ] make `decode_responses=True` as default, so it will smoothly migrate every application


jleveque (approved):
Looks OK to me. Please wait for other reviewers.

tahmed-dev (approved):

PR #95, opened 2020-11-17
[fdbshow] fix fdbshow, show arp commands

dmytroxshevchuk:
Fix issue:
https://github.com/Azure/sonic-buildimage/issues/5927

### DIFF
--- a/src/swsssdk/port_util.py
+++ b/src/swsssdk/port_util.py
@@ -114,7 +114,7 @@ def get_vlan_id_from_bvid(db, bvid):
         Get the Vlan Id from Bridge Vlan Object
     """
     db.connect('ASIC_DB')
-    vlan_obj = db.keys('ASIC_DB', "ASIC_STATE:SAI_OBJECT_TYPE_VLAN:" + bvid)
+    vlan_obj = db.keys('ASIC_DB', str("ASIC_STATE:SAI_OBJECT_TYPE_VLAN:" + bvid))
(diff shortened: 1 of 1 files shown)


prsunny (approved):
@jleveque , can you review and take a look at the issue reported?

jleveque:
@qiluo-msft to comment, as the issue appears to be due to swsssdk change.

bingwang-ms (approved):

dmytroxshevchuk:
@qiluo-msft, please take a look to this PR

prsunny:
@qiluo, this is blocking some tests. Merging for now.

qiluo-msft:
Sorry for late. I don't quite understand why this fix works, could you explain?

bingwang-ms:
I spent some time in investigating this issue, and believe this fix will address the issue correctly.
The root cause for this issue is we use  ```SonicV2Connector``` from ```swsscommon.swsscommon``` in ```fdbshow``` master branch. The ```keys``` method is implemented in C++ and wrapped into a python interface.
```
def keys(self, *args, **kwargs):
        return _swsscommon.SonicV2Connector_keys(self, *args, **kwargs)

```
Therefore, a exception will be raised if we pass an unicode to the interface.
```
TypeError("in method 'SonicV2Connector_keys', argument 3 of type 'char const *'",)
```
So, a conversion to str will address the issue.

PR #97, opened 2020-12-14
Add support for hexists in dbconnector

tonytitus:
dbconnector class doesn't support hexists method today. The hexists method is useful in scripts querying redis DB. This PR adds this support.

### DIFF
--- a/src/swsssdk/interface.py
+++ b/src/swsssdk/interface.py
@@ -298,6 +298,19 @@ class DBInterface(object):
             # redis only supports strings. if any item is set to string 'None', cast it back to the appropriate type.
             return None if val == b'None' else val
 
+
+    @blockable
+    def hexists(self, db_name, _hash, key):
+        """
+        Check if the value of Key %key exists in Hashtable %hash
+        in Database %db_name
+
+        Parameter %blocking indicates whether to wait
+        when the query fails
+        """
+        client = self.redis_clients[db_name]
+        return (client.hexists(_hash, key))
+
     @blockable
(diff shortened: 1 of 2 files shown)


src/swsssdk/interface.py
```diff
@@ -298,6 +298,19 @@ def get(self, db_name, _hash, key):
             # redis only supports strings. if any item is set to string 'None', cast it back to the appropriate type.
             return None if val == b'None' else val
 
+
+    @blockable
+    def hexists(self, db_name, _hash, key):
```
qiluo-msft, line 303: >hexists [](start = 8, length = 7)

swsssdk is going to be deprecate in future. The same functionality is supported by [https://github.com/Azure/sonic-swss-common/blob/master/common/dbinterface.cpp](https://github.com/Azure/sonic-swss-common/blob/master/common/dbinterface.cpp)
and [https://github.com/Azure/sonic-swss-common/blob/master/common/sonicv2connector.cpp](https://github.com/Azure/sonic-swss-common/blob/master/common/sonicv2connector.cpp).
They are implemented in C++, and with a SWIG wrapper, it could be used in python project.

New python code should use sonic-swss-common instead of swsssdk.
Could you move your contribution there?

qiluo-msft (changes requested):
Blocked

qiluo-msft:
Close in favor of https://github.com/Azure/sonic-swss-common/pull/432

qiluo-msft:
Close in favor of https://github.com/Azure/sonic-swss-common/pull/432

PR #99, opened 2020-12-29
Fix bug: use instance variable instead of class variable in ConfigDBConnector

qiluo-msft:
And add instance method `serialize_key` using instance variable.

### DIFF
--- a/src/swsssdk/configdb.py
+++ b/src/swsssdk/configdb.py
@@ -26,8 +26,6 @@ PY3K = sys.version_info >= (3, 0)
 class ConfigDBConnector(SonicV2Connector):
 
     INIT_INDICATOR = 'CONFIG_DB_INITIALIZED'
-    TABLE_NAME_SEPARATOR = '|'
-    KEY_SEPARATOR = '|'
 
     def __init__(self, decode_responses=True, **kwargs):
         # By default, connect to Redis through TCP, which does not requires root.
@@ -49,19 +47,26 @@ class ConfigDBConnector(SonicV2Connector):
            'namespace' is implicitly passed to the parent SonicV2Connector class.
         """
         super(ConfigDBConnector, self).__init__(**kwargs)
+        # Trick: to achieve static/instance method "overload", we must use initize the function in ctor
+        # ref: https://stackoverflow.com/a/28766809/2514803
+        self.serialize_key = self._serialize_key
+        self.deserialize_key = self._deserialize_key
+
+        self.TABLE_NAME_SEPARATOR = '|'
+        self.KEY_SEPARATOR = '|'
         self.handlers = {}
 
     def __wait_for_db_init(self):
         client = self.get_redis_client(self.db_name)
         pubsub = client.pubsub()
-        initialized = client.get(self.INIT_INDICATOR)
+        initialized = client.get(ConfigDBConnector.INIT_INDICATOR)
         if not initialized:
-            pattern = "__keyspace@{}__:{}".format(self.get_dbid(self.db_name), self.INIT_INDICATOR)
+            pattern = "__keyspace@{}__:{}".format(self.get_dbid(self.db_name), ConfigDBConnector.INIT_INDICATOR)
             pubsub.psubscribe(pattern)
             for item in pubsub.listen():
                 if item['type'] == 'pmessage':
                     key = item['channel'].split(':', 1)[1]
-                    if key == self.INIT_INDICATOR:
+                    if key == ConfigDBConnector.INIT_INDICATOR:
                         initialized = client.get(self.INIT_INDICATOR)
                         if initialized:
                             break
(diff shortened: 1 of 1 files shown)


src/swsssdk/configdb.py
```diff
@@ -49,19 +47,26 @@ def __init__(self, decode_responses=True, **kwargs):
            'namespace' is implicitly passed to the parent SonicV2Connector class.
         """
         super(ConfigDBConnector, self).__init__(**kwargs)
+        # Trick: to achieve static/instance method "overload", we must use initize the function in ctor
+        # ref: https://stackoverflow.com/a/28766809/2514803
+        self.serialize_key = self._serialize_key
```
renukamanavalan, line 52: What are we trying to solve here ?
Per link pasted above (Thanks for reference), all we need to declare instance variation for the static methods as suggested -- more a hack. But this code looks not  the same. Can you please help understand?

I guess, (based on  those static methods) the root cause could be Class & instance variable values for *_SEPARATOR could be different. 
Fix: 
In db_connect, you may add the following lines and see if that fixes the issue.

ConfigDBConnector.TABLE_NAME_SEPARATOR = self.TABLE_NAME_SEPARATOR
ConfigDBConnector.KEY_SEPARATOR = self.KEY_SEPARATOR

src/swsssdk/configdb.py
```diff
@@ -49,19 +47,26 @@ def __init__(self, decode_responses=True, **kwargs):
...
+
+        self.TABLE_NAME_SEPARATOR = '|'
+        self.KEY_SEPARATOR = '|'
         self.handlers = {}
 
     def __wait_for_db_init(self):
         client = self.get_redis_client(self.db_name)
         pubsub = client.pubsub()
-        initialized = client.get(self.INIT_INDICATOR)
+        initialized = client.get(ConfigDBConnector.INIT_INDICATOR)
```
renukamanavalan, line 62: It appears to use the class value even with self prefix. Effectively a this code change is no-op, though your update appears good. Guess, ConfigDBCOnnector is soon due for retirement, so we may not need to  polish.

src/swsssdk/configdb.py
```diff
@@ -151,16 +156,32 @@ def typed_to_raw(self, typed_data):
...
         return raw_data
 
+    # Note: we could not use a class variable for KEY_SEPARATOR, but original dependent code is using
+    # these static functions. So we implement both static and instance functions with the same name.
+    # The static function will behave according to ConfigDB separators.
     @staticmethod
     def serialize_key(key):
         if type(key) is tuple:
-            return ConfigDBConnector.KEY_SEPARATOR.join(key)
+            return '|'.join(key)
```
renukamanavalan, line 165: We should not hardcode separator, but use the one obtained upon db_connect,

src/swsssdk/configdb.py
```diff
@@ -151,16 +156,32 @@ def typed_to_raw(self, typed_data):
...
+    def _serialize_key(self, key):
+        if type(key) is tuple:
+            return self.KEY_SEPARATOR.join(key)
         else:
             return str(key)
 
     @staticmethod
     def deserialize_key(key):
-        tokens = key.split(ConfigDBConnector.KEY_SEPARATOR)
+        tokens = key.split('|')
```
renukamanavalan, line 177: Same here. No hard coding separator.

src/swsssdk/configdb.py
qiluo-msft: This is static method, which already used by applications, and they are assuming ConfigDB separator.
Below is an instance method as you expected, and with a trick, I made them the same name.

---
In reply to: [549809761](https://github.com/Azure/sonic-py-swsssdk/pull/99#discussion_r549809761) [](ancestors = 549809761)

src/swsssdk/configdb.py
qiluo-msft: That's my convention, in order not to misuse class variable.

---
In reply to: [549806293](https://github.com/Azure/sonic-py-swsssdk/pull/99#discussion_r549806293) [](ancestors = 549806293)

src/swsssdk/configdb.py
qiluo-msft: Static method `serialize_key` does not make sense because it does not take the instance separator. So I implement the instance method `serialize_key`.
However the statice method is used even before `do_connnect` was introduced. They are assuming separator `|`. So I must keep both static method and instance method to make all applications happy.
It is not straightforward to `overload` this way in python. I must use this trick.


---
In reply to: [549801673](https://github.com/Azure/sonic-py-swsssdk/pull/99#discussion_r549801673) [](ancestors = 549801673)

renukamanavalan: My only concern: Code duplication. Two serialize & de-serialize methods. If the code can be combined, with separator passed in, it will be  better. Or some other more smarter pythonic way.

e.g. 
@static 
serialize(key, sep="|")

def _serialize(self, key)
  ConfigDBConnector,serialize(key, self.KEY_SEPARATOR)

renukamanavalan (approved):

Issue #100, opened 2020-12-31
could sonic-db-cli return JSON string for hgetall ?


Junchao-Mellanox:
Sometimes, sonic-mgmt need check values in redis db by calling "sonic-db-cli $DB HGETALL $KEY". And it would be great if the output of such command can be parsed to a python dictionary easily. 

However, sonic-db-cli now return raw string for hgetall comamnd. For example:

```
admin@r-leopard-01:~$ sonic-db-cli CONFIG_DB HGETALL "TELEMETRY|gnmi"
{'client_auth': 'false', 'log_level': '2', 'port': '50051'}
```

Python json module cannot parse such string to a dictionary:
```
{'client_auth': 'false', 'log_level': '2', 'port': '50051'}
```

I would suggest to change the following code in sonic-db-cli from:

```python
        if resp is None:
            print()
        elif isinstance(resp, list):
            print("\n".join(resp))
        else:
            print(resp)
```

to 

```python
        if resp is None:
            print()
        elif isinstance(resp, list):
            print("\n".join(resp))
        elif isinstance(resp, dict):
            json.dumps(resp)
        else:
            print(resp)
```

Any thought?

Junchao-Mellanox:
As we could use ast.literal_eval to de-serialize dict from string, I am closing this issue.

PR #101, opened 2021-01-18
Optimize sonic-db-cli dict type output to JSON format

Junchao-Mellanox:
Why I did this?

Optimize sonic-db-cli to fix issue https://github.com/Azure/sonic-py-swsssdk/issues/100. 

Sometimes, sonic-mgmt need check values in redis db by calling "sonic-db-cli $DB HGETALL $KEY". And it would be great if the output of such command can be parsed to a python dictionary easily. 

However, sonic-db-cli now return raw string for hgetall comamnd. For example:

```
admin@r-leopard-01:~$ sonic-db-cli CONFIG_DB HGETALL "TELEMETRY|gnmi"
{'client_auth': 'false', 'log_level': '2', 'port': '50051'}
```

Python json module cannot parse such string to a dictionary:
```
{'client_auth': 'false', 'log_level': '2', 'port': '50051'}
```

One way to parse this string to a dictionary is using "eval". If using eval is acceptable in sonic-mgmt code, I can close this PR.

How I did this?

```python
        if resp is None:
            print()
        elif isinstance(resp, list):
            print("\n".join(resp))
        else:
            print(resp)
```

to 

```python
        if resp is None:
            print()
        elif isinstance(resp, list):
            print("\n".join(resp))
        elif isinstance(resp, dict):
            json.dumps(resp)
        else:
            print(resp)
```

How I verify this?

Manual test.

### DIFF
--- a/src/swsssdk/scripts/sonic-db-cli
+++ b/src/swsssdk/scripts/sonic-db-cli
@@ -4,6 +4,7 @@ import sys
 import swsssdk
 import redis
 import argparse
+import json
 from multiprocessing import Pool
 from functools import partial
 
@@ -80,6 +81,8 @@ def execute_cmd(dbname, cmd, namespace, use_unix_socket=False):
             print()
         elif isinstance(resp, list):
             print("\n".join(resp))
+        elif isinstance(resp, dict):
+            print(json.dumps(resp))
         else:
             print(resp)
         sys.exit(0)


liat-grozovik:
@lguohan can you review or assign someone to review?

src/swsssdk/scripts/sonic-db-cli
```diff
@@ -80,6 +81,8 @@ def execute_cmd(dbname, cmd, namespace, use_unix_socket=False):
             print()
         elif isinstance(resp, list):
             print("\n".join(resp))
+        elif isinstance(resp, dict):
+            print(json.dumps(resp))
```
qiluo-msft, line 85: There is a design principle at the birth of this script
> sonic-db-cli output format mimic the non-tty mode output format from redis-cli

In order to easily replace the usage of `redis-cli` with `sonic-db-cli`. The only exception is the `list` type (such as output  of `keys`)
1. `redis-cli` outputs with blanks as delimitator
2. `sonic-db-cli` outputs with '\n' as delimitator
3. Fortunately all the parsing code of the output is not impacted

I see currently the dict type is not well implemented according to the design principle.
```
$ docker exec database redis-cli -n 4 hgetall "PORT|Ethernet20"
admin_status
up
alias
etp6
description
NEIGH06T2:Ethernet1
index
6
lanes
20,21,22,23
mtu
9100
pfc_asym
off
speed
40000
```

You proposed a new feature instead of a bug fix. I suggest add it with new command line options like:
```
$ sonic-db-cli --json CONFIG_DB HGETALL "TELEMETRY|gnmi"
```

Junchao-Mellanox: I agree. But I'd like to drop this PR since we could use ast.literal_eval to de-serialize dict from string.

PR #107, opened 2021-06-15
Add missing functions to be on par with swsscommon.

SuvarnaMeenakshi:
Signed-off-by: Suvarna Meenakshi <sumeenak@microsoft.com>
These functions will be required for unit-test in components which use swsscommon library.

### DIFF
--- a/src/swsssdk/dbconnector.py
+++ b/src/swsssdk/dbconnector.py
@@ -105,6 +105,14 @@ class SonicDBConfig(object):
         SonicDBConfig._sonic_db_config_init = True
 
     @staticmethod
+    def isInit():
+        return SonicDBConfig._sonic_db_config_init
+
+    @staticmethod
+    def initialize(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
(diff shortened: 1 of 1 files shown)


src/swsssdk/dbconnector.py
```diff
@@ -104,6 +104,14 @@ def load_sonic_db_config(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
             raise RuntimeError(msg)
         SonicDBConfig._sonic_db_config_init = True
 
+    @staticmethod
+    def isInit():
+        return SonicDBConfig._sonic_db_config_init
+
+    @staticmethod
+    def initialize(sonic_db_file_path=SONIC_DB_CONFIG_FILE):
```
qiluo-msft, line 112: This function is not needed.
Check https://github.com/Azure/sonic-swss-common/blob/master/common/dbconnector.h#L42-L49. `load_sonic_db_config` will be mapped to `initialize`, so you can always `load_sonic_db_config` in python code no matter which library you are using.

SuvarnaMeenakshi: Removed initialize function

qiluo-msft (approved):

PR #108, opened 2021-06-23
Added loopback interface and changed index for mgmt and portchannel

suresh-rupanagudi:
1)Added loopback interface ifIndex
2)ifIndex for mgmt_port changed from 10000 to 1000.
  As per LLDP-MIB valid range for lldpLocPortId and lldpRemLocalPortNum is (1..4096).
  So if we use MGMT_PORT index as 10000 then it will be out of bounds. So MGMT_PORT index is changed to 1000.
3)ifIndex for port_chanel changed from 1000 to 10000


suresh-rupanagudi:
snmpwalk on ifDescr
IF-MIB::ifDescr.2100 = STRING: Vlan100
IF-MIB::ifDescr.2200 = STRING: Vlan200
IF-MIB::ifDescr.10100 = STRING: PortChannel100
IF-MIB::ifDescr.20000 = STRING: Loopback0

PR #109, opened 2021-06-28
[port_util] Allow system without ports in config db run without errors

liorghub:
Signed-off-by: liora <liora@nvidia.com>

**What I did**
Allow system with no ports in config db to run without errors/warnings. 
This is needed for modular system which should boot and run properly without line cards.

**How I did it**
Check if table exists before calling get_all for it and return empty list if there are no ports in COUNTERS_PORT_NAME_MAP table in counters db.

**How to verify it**
Run snmpwalk on the root oid.

**Important**
This PR must be merged before PR https://github.com/Azure/sonic-snmpagent/pull/221 is merged.

### DIFF
--- a/src/swsssdk/port_util.py
+++ b/src/swsssdk/port_util.py
@@ -61,10 +61,19 @@ def get_interface_oid_map(db):
     """
         Get the Interface names from Counters DB
     """
+    if_name_map = {}
+    if_id_map = {}
+
     db.connect('COUNTERS_DB')
-    if_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_PORT_NAME_MAP', blocking=True)
-    if_lag_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_LAG_NAME_MAP', blocking=True)
-    if_name_map.update(if_lag_name_map)
+    if db.exists('COUNTERS_DB', 'COUNTERS_PORT_NAME_MAP'):
+        if_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_PORT_NAME_MAP', blocking=True)
+
+    if db.exists('COUNTERS_DB', 'COUNTERS_LAG_NAME_MAP'):
+        if_lag_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_LAG_NAME_MAP', blocking=True)
+        if_name_map.update(if_lag_name_map)
+
+    if not if_name_map:
+        return {}, {}
 
     oid_pfx = len("oid:0x")
     if_name_map = {if_name: sai_oid[oid_pfx:] for if_name, sai_oid in if_name_map.items()}
--- a/src/swsssdk/port_util.py
+++ b/src/swsssdk/port_util.py
@@ -57,20 +57,14 @@ def get_index_from_str(if_name):
         if match:
             return int(match.group(1)) + baseidx
 
-def get_interface_oid_map(db):
+def get_interface_oid_map(db, blocking=True):
     """
         Get the Interface names from Counters DB
     """
-    if_name_map = {}
-    if_id_map = {}
-
     db.connect('COUNTERS_DB')
-    if db.exists('COUNTERS_DB', 'COUNTERS_PORT_NAME_MAP'):
-        if_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_PORT_NAME_MAP', blocking=True)
-
-    if db.exists('COUNTERS_DB', 'COUNTERS_LAG_NAME_MAP'):
-        if_lag_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_LAG_NAME_MAP', blocking=True)
-        if_name_map.update(if_lag_name_map)
+    if_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_PORT_NAME_MAP', blocking)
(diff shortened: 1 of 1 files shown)


src/swsssdk/port_util.py
```diff
@@ -61,10 +61,19 @@ def get_interface_oid_map(db):
...
         Get the Interface names from Counters DB
     """
+    if_name_map = {}
+    if_id_map = {}
+
     db.connect('COUNTERS_DB')
-    if_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_PORT_NAME_MAP', blocking=True)
-    if_lag_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_LAG_NAME_MAP', blocking=True)
-    if_name_map.update(if_lag_name_map)
+    if db.exists('COUNTERS_DB', 'COUNTERS_PORT_NAME_MAP'):
```
qiluo-msft, line 68: >exists [](http://example.com/codeflow?start=10&length=6)

You don't need to check `exists()` first. Just call `get_all()` with `blocking=False`, and the result is either empty or non-empty.

qiluo-msft:
> def get_interface_oid_map(db):

What if a user really want to block waiting for ports to appear? Can we set `blocking=True` as default paramter so application could migrate behavior incrementally?

---
Refers to: src/swsssdk/port_util.py:60 in 40637b8. [](commit_id = 40637b8d2b5a6c2b179b8e6b45d241e1bb843c92, deletion_comment = False)

qiluo-msft (changes requested):
As comments

src/swsssdk/port_util.py
liorghub: Done, thanks.

liorghub:
> > def get_interface_oid_map(db):
> 
> What if a user really want to block waiting for ports to appear? Can we set `blocking=True` as default paramter so application could migrate behavior incrementally?
> 
> Refers to: src/swsssdk/port_util.py:60 in [40637b8](https://github.com/Azure/sonic-py-swsssdk/commit/40637b8d2b5a6c2b179b8e6b45d241e1bb843c92). [](commit_id = [40637b8](https://github.com/Azure/sonic-py-swsssdk/commit/40637b8d2b5a6c2b179b8e6b45d241e1bb843c92), deletion_comment = False)

Done.

liorghub:
The unitest for this change is in PR https://github.com/Azure/sonic-snmpagent/pull/221

liorghub:
@qiluo-msft can you please approve this commit?
Only when this commit is in we can merge PR https://github.com/Azure/sonic-snmpagent/pull/221.

src/swsssdk/port_util.py
```diff
@@ -57,15 +57,18 @@ def get_index_from_str(if_name):
...
         Get the Interface names from Counters DB
     """
     db.connect('COUNTERS_DB')
-    if_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_PORT_NAME_MAP', blocking=True)
-    if_lag_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_LAG_NAME_MAP', blocking=True)
+    if_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_PORT_NAME_MAP', blocking)
+    if_lag_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_LAG_NAME_MAP', blocking)
     if_name_map.update(if_lag_name_map)
 
+    if if_name_map == {}:
```
qiluo-msft, line 69:
```suggestion
    if if_name_map:
```

liorghub: The code written in lines 69 and 70 is an enhancement. 
If we run on a system without ports (where if_name_map={}) there is no point to execute the rest of the function. 
In line 69 we check if the dictionary is empty and if so we return.
The proposed change is simply not doing what we want to do (it is not checking that the dictionary is empty).

qiluo-msft: Sorry for misleading. My suggestion is actually mistakenly written. I updated.
I totally understand your point. My comment is suggestion to treat empty and None the same way, so it will tolerate the implementation detail of `db.get_all`

liorghub: Done.

src/swsssdk/port_util.py
```diff
@@ -57,15 +57,18 @@ def get_index_from_str(if_name):
...
 
-def get_interface_oid_map(db):
+def get_interface_oid_map(db, blocking=True):
     """
         Get the Interface names from Counters DB
     """
     db.connect('COUNTERS_DB')
-    if_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_PORT_NAME_MAP', blocking=True)
-    if_lag_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_LAG_NAME_MAP', blocking=True)
+    if_name_map = db.get_all('COUNTERS_DB', 'COUNTERS_PORT_NAME_MAP', blocking)
```
mlok-nokia, line 65: Line 65,and 66, function get_all()  cannot accept the non-keyword " ```blocking``` as a positional  argument.  It needs to be as keyword argument ```blocking=True````

liorghub:
@mlok-nokia
Thanks Marty! I have submitted PR https://github.com/Azure/sonic-py-swsssdk/pull/114 to fix this issue.

PR #110, opened 2021-07-07
[voq][chassis] Fix for issue in chassis redis server connection check

vganesan-nokia:
During database docker start, when the redis servers are started, ping pong is done to check the server connection status. For server within the card ping pong is done using unix socket. In VOQ chassis, to check the connection to redis server in supervisor card, we need to use host name and port instead of using unix socket.

### DIFF
--- a/src/swsssdk/scripts/sonic-db-cli
+++ b/src/swsssdk/scripts/sonic-db-cli
@@ -9,7 +9,7 @@ from functools import partial
 
 def handle_single_instance(op, use_unix_socket, inst_info):
     inst_hostname = inst_info['hostname']
-    if use_unix_socket:
+    if use_unix_socket and inst_hostname != 'redis_chassis.server':
         inst_unix_socket_path = inst_info['unix_socket_path']
         r = redis.Redis(host=inst_hostname, unix_socket_path=inst_unix_socket_path)
     else:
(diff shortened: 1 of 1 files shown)


src/swsssdk/scripts/sonic-db-cli
```diff
@@ -9,15 +9,15 @@ from functools import partial
 
 def handle_single_instance(op, use_unix_socket, inst_info):
     inst_hostname = inst_info['hostname']
-    if use_unix_socket:
+    if use_unix_socket and inst_hostname != 'redis_chassis.server':
```
judyjoseph, line 12: The unix socket is used only if we specify it explicitly to sonic-db-cli command, or if --namespace option is passed as argument when sonic-db-cli is called. We should not be passing namespace when trying to connect to chassis DB in super visor as it is running in the host in supervisor.

Can you share where exactly it is failing, that we need this condition ?

vganesan-nokia: During database initialization, attempt is made to check the connectivity with redis servers using PING. In the place (database.sh) where SONIC_DB_CLI is used to issue the PING command, "SONIC_DB_CLI" is defined as "sonic-db-cli -n <namespace>" during the initialization of asic specific databsae. (this happens in the linecard for all asics). Since the PING command attempts to ping to all the instances in the database config which includes chassis redis server and since the name space is passed (as mentioned above), the "use_unix_socket = true" is used for all the instance including chassis redis server which is in supervisor card.

judyjoseph (approved):

PR #115, opened 2021-11-08
Add option to suppress warnings when querying Redis

TACappleman:
In some circumstances calls are made to query the Redis database without checking if a given key (or key/field pair) exists, and with code to handle the case where it does not. For example many show commands query all possible relevant data, without checking if it exists. In these cases it would be useful to be able to prevent warning logs being written for every missing field, and so we add that option in this PR.

### DIFF
--- a/src/swsssdk/interface.py
+++ b/src/swsssdk/interface.py
@@ -280,7 +280,7 @@ class DBInterface(object):
             return keys
 
     @blockable
-    def get(self, db_name, _hash, key):
+    def get(self, db_name, _hash, key, log_warning=True):
         """
         Retrieve the value of Key %key from Hashtable %hash
         in Database %db_name
@@ -292,7 +292,8 @@ class DBInterface(object):
         val = client.hget(_hash, key)
         if not val:
             message = "Key '{}' field '{}' unavailable in database '{}'".format(_hash, key, db_name)
-            logger.warning(message)
+            if log_warning:
+                logger.warning(message)
             raise UnavailableDataError(message, _hash)
         else:
             # redis only supports strings. if any item is set to string 'None', cast it back to the appropriate type.
@@ -310,7 +311,8 @@ class DBInterface(object):
         table = client.hgetall(_hash)
         if not table:
             message = "Key '{}' unavailable in database '{}'".format(_hash, db_name)
-            logger.warning(message)
+            if log_warning:
+                logger.warning(message)
             raise UnavailableDataError(message, _hash)
         else:
             # redis only supports strings. if any item is set to string 'None', cast it back to the appropriate type.
--- a/src/swsssdk/interface.py
+++ b/src/swsssdk/interface.py
@@ -40,6 +40,7 @@ def blockable(f):
                 return ret_data
             except UnavailableDataError as e:
                 if blocking:
+                	logger.warning(e.message)
                     if db_name in inst.keyspace_notification_channels:
                         result = inst._unavailable_data_handler(db_name, e.data)
                         if result:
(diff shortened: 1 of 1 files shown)


TACappleman:
@qiluo-msft could you take a look at this PR please (looks like you've been the most recent approver in this PR)?

This is a very minor utility change that will be useful in preventing unnecessary warning logs in certain scenarios - e.g. when show commands query Redis for many fields without checking if they're present, and are able to handle the case where they're not present.

src/swsssdk/interface.py
```diff
@@ -280,7 +280,7 @@ def keys(self, db_name, pattern='*'):
             return keys
 
     @blockable
```
qiluo-msft, line 282: >blockable [](http://example.com/codeflow?start=5&length=9)

We are deprecate the python implementation and in favor of swss-common (SWIG powered). Would you like to change in that repo at the same time or only in that repo?

src/swsssdk/interface.py
```diff
@@ -280,7 +280,7 @@ def keys(self, db_name, pattern='*'):
...
-    def get(self, db_name, _hash, key):
+    def get(self, db_name, _hash, key, log_warning=True):
```
qiluo-msft, line 283: >log_warning [](http://example.com/codeflow?start=39&length=11)

name it as `verbose` ?

src/swsssdk/interface.py
TACappleman: Made another PR for that here: https://github.com/Azure/sonic-swss-common/pull/566

src/swsssdk/interface.py
TACappleman: Done

qiluo-msft:
>                 if blocking:

I have second thought about the solution. Since the functions are throwing exceptions with the message, it is the caller's responsibility to handle and log

0. We should remove the logging statements from the basic functions `keys`/`get`/`get_all`/etc, if it is followed by a throw.
1. We may add the logging statements into the `blocking` branch in the function `wrapped`. Then no behavior change in this condition.
2. For non blocking branch, we swallow the `UnavailableDataError` and `return None`. The caller should get well understanding what happening, and even the caller does not handle well, the further exception on None object should be clear. So I think we are okay with the behavior change.

What do you think?




---
Refers to: src/swsssdk/interface.py:42 in b8838a8. [](commit_id = b8838a8290a7bc0bdebbc455196f7faed3b7d819, deletion_comment = False)

TACappleman:
> > ```
> >             if blocking:
> > ```
> 
> I have second thought about the solution. Since the functions are throwing exceptions with the message, it is the caller's responsibility to handle and log
> 
> 0. We should remove the logging statements from the basic functions `keys`/`get`/`get_all`/etc, if it is followed by a throw.
> 1. We may add the logging statements into the `blocking` branch in the function `wrapped`. Then no behavior change in this condition.
> 2. For non blocking branch, we swallow the `UnavailableDataError` and `return None`. The caller should get well understanding what happening, and even the caller does not handle well, the further exception on None object should be clear. So I think we are okay with the behavior change.
> 
> What do you think?
> 
> Refers to: src/swsssdk/interface.py:42 in [b8838a8](https://github.com/Azure/sonic-py-swsssdk/commit/b8838a8290a7bc0bdebbc455196f7faed3b7d819). [](commit_id = [b8838a8](https://github.com/Azure/sonic-py-swsssdk/commit/b8838a8290a7bc0bdebbc455196f7faed3b7d819), deletion_comment = False)

Works for me. I've updated this and the swss-common equivalent: https://github.com/Azure/sonic-swss-common/pull/566

qiluo-msft (approved):

PR #113, opened 2021-07-22
[Ethernet-IB][index] Modified the port_util to support the VoQ Inband …

mlok-nokia:
Ethernet-IB port is introduced for the VoQ chassis. This requires to define the index of the Ethernet-IB to allow the application to get the index from the port name. Follow the existing design, define the Inband port index base as 11000.

PR #118, opened 2022-03-01
define index for VoQ recirc port

ysmanman:
VoQ inband port index was added in https://github.com/Azure/sonic-py-swsssdk/pull/113. This PR is to define index for VoQ recirc port.


arlakshm (approved):
lgtm

SuvarnaMeenakshi (approved):

PR #128, opened 2022-11-01
Fix UT test_BlockUseSwsssdk()

liuh-80:
#### Why I did it
Fix this issue: https://github.com/sonic-net/sonic-buildimage/issues/12558
UT failed under python2: test_BlockUseSwsssdk()
>       result = subprocess.run(["python", "-c", "import swsssdk;exit()"], capture_output=True, cwd=swsssdk_path)
E       AttributeError: 'module' object has no attribute 'run'

#### How I did it
Update UT to use subprocess.check_output(), which support by both python3 and python2.

#### How to verify it
Pass all UT.

#### Description for the changelog
Fix UT test_BlockUseSwsssdk()

### DIFF
--- a/test/test_moduleLoad.py
+++ b/test/test_moduleLoad.py
@@ -42,6 +42,6 @@ def test_BlockUseSwsssdk():
     try:
         subprocess.check_output(["python", "-c", "import swsssdk;exit()"], stderr=subprocess.STDOUT, cwd=swsssdk_path)
     except subprocess.CalledProcessError as e:
-        result = e.output
+        result = e.output.decode("utf-8")
 
     assert "deprecated" in result
--- a/test/test_moduleLoad.py
+++ b/test/test_moduleLoad.py
@@ -38,6 +38,6 @@ class Test_load_sonic_db_config(TestCase):
 def test_BlockUseSwsssdk():
     # Import swsssdk will throw exception with deprecated message.
     swsssdk_path = os.path.join(modules_path, 'src')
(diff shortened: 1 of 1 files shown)


liuh-80:
Tested with this PR: https://github.com/sonic-net/sonic-buildimage/pull/12565

test/test_moduleLoad.py
```diff
@@ -38,6 +38,15 @@ def test__dbConfig(self):
 def test_BlockUseSwsssdk():
     # Import swsssdk will throw exception with deprecated message.
     swsssdk_path = os.path.join(modules_path, 'src')
-    result = subprocess.run(["python", "-c", "import swsssdk;exit()"], capture_output=True, cwd=swsssdk_path)
+    result = None
+    python_command = "python"
+    
+    if sys.version_info.major == 3:
```
qiluo-msft, line 44: >major [](http://example.com/codeflow?start=24&length=5)

There is a package "python-is-python3", which will symlink python to python3.

You assume python==python2.x, it will not be true if above package exist.

Is it possible to `import swsssdk` in this function `test_BlockUseSwsssdk`?

liuh-80: Can't import swsssdk in test case, because if we import it, the code in swsssdk will know it's running inside UT. then will not throw exception.

The version check is to fix the 'python' command does not exist issue. so, if "python-is-python3" installed, this UT still can work.

qiluo-msft (approved):

PR #129, opened 2022-11-09
Remove LOGLEVEL DB since is no longer used

EdenGri:
This PR is part of the following HLD:
Persistent loglevel HLD: https://github.com/sonic-net/SONiC/pull/1041
**- What I did**
Deleted the LOGLEVEL_DB.
After the Logger tables moved from the LOGLEVEL_DB to the CONFIG_DB and the jinja2_cache was deleted the LOGLEVEL_DB is not in use.
**- How I did it**
Removed the LOGLEVEL_DB from the SONiC code

**- How to verify it**
All tests were passed

### DIFF
--- a/src/swsssdk/config/database_config.json
+++ b/src/swsssdk/config/database_config.json
@@ -22,11 +22,6 @@
             "separator": ":",
             "instance" : "redis"
         },
-        "LOGLEVEL_DB" : {
-            "id" : 3,
-            "separator": ":",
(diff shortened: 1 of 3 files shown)


dprital (dismissed):

qiluo-msft (approved):

qiluo-msft:
sonic-py-swsssdk is deprecated and only used in unit test of other repos. The coverage checker is safe to override.

PR #135, opened 2023-04-09
[Security] Fix the redis security issue CVE-2023-28858 and  CVE-2023-28859

xumia:
[Security] Fix the redis security issue CVE-2023-28858 and  CVE-2023-28859

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,7 @@
 from setuptools import setup
 
 dependencies = [
-    'redis==4.3.6;python_version >= "3.0"',
+    'redis==4.5.4;python_version >= "3.0"',
     'redis>=3.5.3;python_version < "3.0"',
     'redis-dump-load',
 ]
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -1,7 +1,8 @@
 from setuptools import setup
 
 dependencies = [
-    'redis==3.5.3',
+    'redis==4.5.4;python_version >= "3.0"',
```
qiluo-msft, line 4:
```suggestion
    'redis>=4.5.4;python_version >= "3.0"',
```
The true versions are specified by Dockerfile, such as dockers/docker-config-engine-bullseye/Dockerfile.j2. It may use later versions, we do not want setup.py to install an older version in that case.

xumia: It is a good suggestion, I will send a PR to fix the master branch.

qiluo-msft (approved):

qiluo-msft (approved):

PR #140, opened 2023-05-26
Allow deleting tables before updating

seiferteric:
**What is this**
There is a use-case when you want to update a table, but delete all entries in that table beforehand, then update in a single transaction. A use-case is where we collect stats on all processes and docker containers periodically. In this case, process and containers may com and go, and so you would want to delete all the old entries and add back the new ones to the table each time instead of figuring out which processes had gone and which have stayed each time.

The issue with the current implementation is that since you must pass the table name as "None" to delete, but then I must also pass the table data, but I can't do that because it's the same table. So my data would have to be this for example `{"PROC_STATS": None, "PROC_STATS": {...actual data...}}` which is invalid because of the duplicate key. The solution here is the pass a list of table names to delete before doing the update, but as part of the same transaction.


seiferteric:
@qiluo-msft can you review? Thanks.

